PR #10, opened 2023-11-28
Added new method for firebase directly login

akaboshinit:
Thanks for developing a great package!

We have come up with an implementation that makes initialization easier and more flexible.

When I try to use this package on [dart_edge](https://github.com/invertase/dart_edge)
- `Credential.fromServiceAccount` requires access to a local file, so on edge I get an error because I can't access the file
- `Credential.fromApplicationDefaultCredentials` is cumbersome to set the environment because the edge environment may not be a GCP environment.

In this pull request, Credential.fromServiceAccountParams is used to set the credentials directly from the ServiceAccount string.

example:
```dart
final admin = FirebaseAdminApp.initializeApp(
    'project-id',
    Credential.fromServiceAccountParams(
      clientId: "123*******",
      privateKey: "-----BEGIN PRIVATE KEY-----\n**********",
      clientEmail: "******@******.iam.gserviceaccount.com",
));
```


packages/dart_firebase_admin/lib/src/app/credential.dart
```diff
@@ -25,6 +25,22 @@ class Credential {
     return Credential._(serviceAccountCredentials);
   }
 
+  /// Log in to firebase from a service account file parameters.
+  factory Credential.fromServiceAccountParams({
+    required String clientId,
+    required String privateKey,
+    required String clientEmail,
```
rrousselGit, line 32:
```suggestion
    required String email,
```

packages/dart_firebase_admin/lib/src/app/credential.dart
```diff
@@ -25,6 +25,22 @@ class Credential {
...
+  }) {
+    final serviceAccountCredentials = auth.ServiceAccountCredentials.fromJson({
+      'type': 'service_account',
+      'private_key': privateKey,
+      'client_email': clientEmail,
+      'client_id': clientId,
+    });
```
rrousselGit, line 39: No need for fromJSON:

```suggestion
    final serviceAccountCredentials = auth.ServiceAccountCredentials(
      email,
      clientId,
      privateKey,
    );
```

PR #14, opened 2023-11-29
fix: proper conversion of Uint8List

samtuga1:


rrousselGit:
I see I'm using an SDK that's up-to-date: https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#dartconvert

I'll fix this by bumping the minimum Dart SDK instead.

rrousselGit:
Thanks for the fix though!

samtuga1:
> I don't see why your changes should change anything. `utf8.encode` returns a `Uint8List` for me, not `List<int>`.
> 
> Out of curiosity, which Dart version are you using?

I am using Dart SDK version: 3.1.0 (stable)

samtuga1:
> I see I'm using an SDK that's up-to-date: https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#dartconvert
> 
> I'll fix this by bumping the minimum Dart SDK instead.

That will certainly work as well 👍🏽

PR #19, opened 2024-02-02
Support custom emulator hosts and ports

nullbtb:
I have multiple projects with different emulator ports so I thought it would be beneficial to be able to configure this.

### DIFF
--- a/packages/dart_firebase_admin/lib/src/app/firebase_admin.dart
+++ b/packages/dart_firebase_admin/lib/src/app/firebase_admin.dart
@@ -1,13 +1,32 @@
 part of '../app.dart';
 
 class FirebaseAdminApp {
-  FirebaseAdminApp.initializeApp(this.projectId, this.credential);
+  FirebaseAdminApp.initializeApp(
+    this.projectId,
+    this.credential, {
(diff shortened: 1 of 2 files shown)


rrousselGit:
These probably should be picked up from the `firebase.json` instead of manually passed

nullbtb:
Hmm, I think that could be interesting, but this would introduce a dependency on the firebase.json file.  This would be undesirable because now this package would have to be given a path, validate file contents, and so on.  In addition if their schema changes, it would break prior versions of this package.   In general I think it's preferable to handle it like this and let the user worry about that aspect.  This provides them more flexibility for testing, for placing the firebase.json in a different project , etc.

Ehesp:
We should follow standard practice here, and detect an env variable:

```
export FIREBASE_AUTH_EMULATOR_HOST="127.0.0.1:9099"
```

See: https://firebase.google.com/docs/emulator-suite/connect_auth#admin_sdks

Appreciate the PR, however we should stick with the same process other admin SDKs follow.

nullbtb:
Sure that makes sense.  I didn't realize they have different standards for admin sdks and client sdks.  I'll close this PR then.  Thanks guys!

PR #27, opened 2024-03-08
feat: Update Timestamp Constructor with nanoseconds

KKimj:
Hello!

This update enables preservation of nanoseconds below 1000, addressing the issue where fractions of milliseconds were being discarded.
image

Thanks!


rrousselGit (changes requested):
There's some value to it.

Could you add tests for this constructor? Since it's now public.
And add dartdocs for it too.

KKimj:
@rrousselGit 

Hello! : )
I've added some docstrings and test code for `Timestamp`.
Please let me know if you have any feedback. Thank you!

KKimj:
Ah, and I followed the docstring of the code below! :)

https://github.com/firebase/flutterfire/blob/master/packages/cloud_firestore/cloud_firestore_platform_interface/lib/src/timestamp.dart#L17C1-L24C51

rrousselGit (approved):

packages/dart_firebase_admin/lib/src/google_cloud_firestore/timestamp.dart
```diff
@@ -81,8 +91,29 @@ final class Timestamp implements _Serializable {
...
+  ///
+  /// documentRef.set({ 'startTime': Timestamp.fromMicros(42) });
+  /// ```
+  ///
+  /// - [microseconds]: Number of microseconds since Unix epoch
+  /// 1970-01-01T00:00:00Z.
+  ///
+  /// Returns a new [Timestamp] representing the same point in time
+  /// as the given number of microseconds.
+  factory Timestamp.fromMicros(int microseconds) {
```
rrousselGit, line 112: This one needs a test too

KKimj: Done! Thanks : )

Ehesp: Is this constructor necessary? It makes sense to stick as close to the JavaScript implementation as possible imo, and I don't believe this exists there (could be wrong)

KKimj: I believe the Timestamp.fromMicros constructor is necessary. Considering the Dart native environment, the lack of this constructor would result in the loss of microsecond precision.

Additionally, while the microsecondsSinceEpoch property of the Dart DateTime API is available in the web environment, the limitations of JavaScript mean that there will still be a loss of precision in the microsecond unit.

Even though a code like 
```dart
Timestamp.fromMicros(DateTime.now().microsecondsSinceEpoch);
```
would not throw an error

Ehesp: Interesting, thanks for the insight.  I'd much prefer the name to be `fromMicroseconds`, but leave that one to Remi to call :D

rrousselGit (approved):
LGTM thanks!

PR #28, opened 2024-03-21
chore: Migrating to melos >= 3

erickzanardo:
Small PR to migrate the project structure to the one needed to use on melos >=3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,4 +25,4 @@ pubspec.lock
 
 service-account.json
 
-pubspec_overrides.yaml
+**/pubspec_overrides.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,5 @@ pubspec.lock
 .last_build_id
 
 service-account.json
+
+pubspec_overrides.yaml
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,4 +1,5 @@
 name: dart_firebase_admin_workspace
+publish_to: none
 
 environment:
   sdk: '>=3.0.0 <4.0.0'
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,0 +1,6 @@
+name: dart_firebase_admin_workspace
+
+environment:
+  sdk: '>=3.0.0 <4.0.0'
+dev_dependencies:
+  melos: ^5.2.1


.gitignore
```diff
@@ -23,4 +23,6 @@ pubspec.lock
 .settings
 .last_build_id
 
-service-account.json
\ No newline at end of file
+service-account.json
+
+pubspec_overrides.yaml
```
spydon, line 28:
```suggestion
**/pubspec_overrides.yaml
```
It should ignore all pubspec overrides files right?

rrousselGit:
I don't want to use v3 or above in my projects, because I don't want to have to define a root-level pubspec.

erickzanardo:
> I don't want to use v3 or above in my projects, because I don't want to have to define a root-level pubspec.

But given that that is the path that melos is going, shouldn't we just keep up with it? I made this first PR since I am planing on sending a future contribution and I have multiple projects using melos, I feel it is more troublesome to keep changing my melos install per project than to have a top level pubspec.

Unless there is a way of keeping multiple versions of melos in my computer that I am unfamiliar wiht?

rrousselGit:
I'd rather not use melos than having to specify a root `pubspec.yaml`.

If you want to contribute something here, feel free to remove melos entirely (in favour if plain pubspec_overrides.yaml)

spydon:
> I'd rather not use melos than having to specify a root `pubspec.yaml`.
> 
> If you want to contribute something here, feel free to remove melos entirely (in favour if plain pubspec_overrides.yaml)

Now I'm curious, what downsides do you see with having a root-level pubspec?

rrousselGit:
It's invasive. I want projects to be as clean as possible from config files.

Dart is really good at having very little config. I want to keep it this way.

Salakar:
We should **not** be gating contributions over personal preference here, the standard on most Invertase repositories is to use Melos and the root pubspec layout, it has served us well at scale when there's a focus on multiple contributors, rather than ones own preferences.

pubspec.yaml
```diff
@@ -0,0 +1,6 @@
+name: dart_firebase_admin_workspace
+
```
Salakar, line 2: Can we add publish to none here?

erickzanardo: For sure. Added!

Salakar (approved):

PR #32, opened 2024-04-30
feat: Use GOOGLE_APPLICATION_CREDENTIALS if json value

codekeyz:
Modified `Credential.fromApplicationDefaultCredentials` to use `GOOGLE_APPLICATION_CREDENTIALS` from environmental variables if value is `JSON`.

Ideally, this should have been in [googleapis_auth](https://github.com/google/googleapis.dart/blob/bf977ccd6346b06ee8ec53122ff765d86c32acc4/googleapis_auth/lib/auth_io.dart#L46-L67)

### DIFF
--- a/packages/dart_firebase_admin/lib/src/app.dart
+++ b/packages/dart_firebase_admin/lib/src/app.dart
@@ -8,7 +8,6 @@ import 'package:firebaseapis/identitytoolkit/v3.dart' as auth3;
 import 'package:googleapis_auth/auth_io.dart' as auth;
 import 'package:googleapis_auth/googleapis_auth.dart';
 import 'package:meta/meta.dart';
-import 'package:platform/platform.dart';
 
 part 'app/credential.dart';
 part 'app/exception.dart';
--- a/packages/dart_firebase_admin/lib/src/app.dart
+++ b/packages/dart_firebase_admin/lib/src/app.dart
@@ -8,6 +8,7 @@ import 'package:firebaseapis/identitytoolkit/v3.dart' as auth3;
 import 'package:googleapis_auth/auth_io.dart' as auth;
 import 'package:googleapis_auth/googleapis_auth.dart';
 import 'package:meta/meta.dart';
+import 'package:platform/platform.dart';
 
 part 'app/credential.dart';
 part 'app/exception.dart';
--- a/packages/dart_firebase_admin/lib/src/app/credential.dart
+++ b/packages/dart_firebase_admin/lib/src/app/credential.dart
@@ -1,7 +1,6 @@
 part of '../app.dart';
 
 @internal
-@visibleForTesting
 const envSymbol = #_envSymbol;
 
 /// Authentication information for Firebase Admin SDK.
--- a/packages/dart_firebase_admin/lib/src/app/credential.dart
+++ b/packages/dart_firebase_admin/lib/src/app/credential.dart
@@ -1,6 +1,8 @@
 part of '../app.dart';
 
-Platform platform = const LocalPlatform();
+@internal
+@visibleForTesting
+const envSymbol = #_envSymbol;
 
 /// Authentication information for Firebase Admin SDK.
 class Credential {
@@ -47,14 +49,17 @@ class Credential {
     String? serviceAccountId,
   }) {
     ServiceAccountCredentials? creds;
-    final maybeConfig = platform.environment['GOOGLE_APPLICATION_CREDENTIALS'];
+
+    final env =
+        Zone.current[envSymbol] as Map<String, String>? ?? Platform.environment;
+    final maybeConfig = env['GOOGLE_APPLICATION_CREDENTIALS'];
     if (maybeConfig != null) {
       try {
         final decodedValue = jsonDecode(maybeConfig);
         if (decodedValue is Map) {
           creds = ServiceAccountCredentials.fromJson(decodedValue);
         }
-      } catch (_) {}
+      } on FormatException catch (_) {}
     }
 
     return Credential._(
--- a/packages/dart_firebase_admin/lib/src/app/credential.dart
+++ b/packages/dart_firebase_admin/lib/src/app/credential.dart
@@ -46,8 +46,10 @@ class Credential {
   }) {
     ServiceAccountCredentials? creds;
     final maybeConfig = Platform.environment['GOOGLE_APPLICATION_CREDENTIALS'];
-    if (maybeConfig != null && maybeConfig.startsWith('{')) {
-      creds = ServiceAccountCredentials.fromJson(jsonDecode(maybeConfig));
+    if (maybeConfig != null) {
+      try {
+        creds = ServiceAccountCredentials.fromJson(jsonDecode(maybeConfig));
+      } catch (_) {}
     }
 
     return Credential._(
--- a/packages/dart_firebase_admin/lib/src/app/credential.dart
+++ b/packages/dart_firebase_admin/lib/src/app/credential.dart
@@ -48,7 +48,10 @@ class Credential {
     final maybeConfig = Platform.environment['GOOGLE_APPLICATION_CREDENTIALS'];
     if (maybeConfig != null) {
       try {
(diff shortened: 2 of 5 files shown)


packages/dart_firebase_admin/lib/src/app/credential.dart
```diff
@@ -41,11 +41,20 @@ class Credential {
...
-      : this._(
-          null,
-          serviceAccountId: serviceAccountId,
-        );
+  factory Credential.fromApplicationDefaultCredentials({
+    String? serviceAccountId,
+  }) {
+    ServiceAccountCredentials? creds;
+    final maybeConfig = Platform.environment['GOOGLE_APPLICATION_CREDENTIALS'];
+    if (maybeConfig != null && maybeConfig.startsWith('{')) {
```
rrousselGit, line 49: I don't like this random `startWith('{`

I'd rather use `try { jsonDecode } on FormatException {...}`

rrousselGit:
What is this about already?

packages/dart_firebase_admin/lib/src/app/credential.dart
codekeyz: Alright, good suggestion.

Ehesp: Agree, although that's the implementation in node sdk so probably taken from there

codekeyz:
> What is this about already?

@rrousselGit This modifies `Credential.fromApplicationDefaultCredentials` to be aware of instances where `GOOGLE_APPLICATION_CREDENTIALS` value is JSON and not a file path.

lesnitsky (changes requested):

packages/dart_firebase_admin/lib/src/app/credential.dart
```diff
@@ -41,11 +41,22 @@ class Credential {
...
-          serviceAccountId: serviceAccountId,
-        );
+  factory Credential.fromApplicationDefaultCredentials({
+    String? serviceAccountId,
+  }) {
+    ServiceAccountCredentials? creds;
+    final maybeConfig = Platform.environment['GOOGLE_APPLICATION_CREDENTIALS'];
+    if (maybeConfig != null) {
+      try {
+        creds = ServiceAccountCredentials.fromJson(jsonDecode(maybeConfig));
```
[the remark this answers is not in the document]
codekeyz: Yeah, but I don't think we need to throw error here. 

This is specifically checking for valid json value so we can use it, if not, do nothing.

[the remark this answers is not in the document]
lesnitsky: @codekeyz would the underlying `googleapis_auth` throw a meaningful error if provided path is invalid or service account file is not located at a given path?

rrousselGit:
Could you add a test for this? That's quite important logic

packages/dart_firebase_admin/lib/src/app/credential.dart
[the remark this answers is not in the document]
codekeyz: It does. 
```dart
 try {
    credentials = json.decode(await file.readAsString());
  } on IOException {
    throw Exception(
      'Failed to read credentials file from $fileSource',
    );
  } on FormatException {
    throw Exception(
      'Failed to parse JSON from credentials file from $fileSource',
    );
  }
```

codekeyz:
@rrousselGit Please review when you can.

packages/dart_firebase_admin/lib/src/app/credential.dart
```diff
@@ -41,11 +43,25 @@ class Credential {
...
+  }) {
+    ServiceAccountCredentials? creds;
+    final maybeConfig = platform.environment['GOOGLE_APPLICATION_CREDENTIALS'];
+    if (maybeConfig != null) {
+      try {
+        final decodedValue = jsonDecode(maybeConfig);
+        if (decodedValue is Map) {
+          creds = ServiceAccountCredentials.fromJson(decodedValue);
+        }
+      } catch (_) {}
```
rrousselGit, line 57: We shouldn't catch _anything_.
Only FormatExceptions.

packages/dart_firebase_admin/lib/src/app.dart
```diff
@@ -8,6 +8,7 @@ import 'package:firebaseapis/identitytoolkit/v3.dart' as auth3;
 import 'package:googleapis_auth/auth_io.dart' as auth;
 import 'package:googleapis_auth/googleapis_auth.dart';
 import 'package:meta/meta.dart';
+import 'package:platform/platform.dart';
```
rrousselGit, line 11: I don't like adding a new dependency just for this.

packages/dart_firebase_admin/lib/src/app/credential.dart
```diff
@@ -1,5 +1,7 @@
 part of '../app.dart';
 
+Platform platform = const LocalPlatform();
```
rrousselGit, line 3: This shouldn't be public. At the very least, it needs to be `@internal` and the `lib/` export needs to `hide` it.

packages/dart_firebase_admin/test/credential_test.dart
```diff
@@ -66,5 +67,40 @@ void main() {
         Credential.fromServiceAccount(fs.file('service-account.json'));
       });
     });
+
+    group('fromApplicationDefaultCredentials', () {
+      test(
+          'completes if `GOOGLE_APPLICATION_CREDENTIALS` environment-variable is valid service account JSON',
+          () {
+        platform = FakePlatform(
```
rrousselGit, line 75: Tests that modify global state should have an addTearDown that reset the global state,

As mentioned before, I'd rather not have the global variable. Maybe use Zones?
But if we stick with this, we should remove add an addTearDown (and probably create a test util to both set the var and do the tear-down)

rrousselGit:
Thanks for the tests!

packages/dart_firebase_admin/lib/src/app.dart
rrousselGit: One way to do this would be to use zones.

You could do:

```dart
@internal
const envSymbol = #_envSymbol;

...

final env = (Zone.current[envSymbol] ?? Platform.current.environment) as Map<String, String>;
final credential = env['GOOGLE_APPLICATION_CREDENTIALS'];


...

test(..., () {
  runZoned(() => ..., zoneValues: {envSymbol: { 'GOOGLE_APPLICATION_CREDENTIALS': 'foo', );
```

packages/dart_firebase_admin/lib/src/app/credential.dart
codekeyz: Resolved.

packages/dart_firebase_admin/lib/src/app.dart
codekeyz: TIL. Thanks @rrousselGit

packages/dart_firebase_admin/lib/src/app/credential.dart
codekeyz: Removed.

packages/dart_firebase_admin/test/credential_test.dart
codekeyz: Yeah, we're using Zones now

packages/dart_firebase_admin/lib/src/app/credential.dart
```diff
@@ -1,5 +1,9 @@
 part of '../app.dart';
 
+@internal
+@visibleForTesting
```
rrousselGit, line 4:
```suggestion
```

packages/dart_firebase_admin/lib/src/app/credential.dart
```diff
@@ -1,5 +1,8 @@
 part of '../app.dart';
 
+@internal
+const envSymbol = #_envSymbol;
```
rrousselGit, line 4: This needs to be hidden by the associated export, as we don't want this to be part of the public API

rrousselGit (approved):
LGTM once `envSymbol` is hidden

lesnitsky (approved):

PR #35, opened 2024-06-07
Able to assign `host` and `port` for emualtor

hn12404988:
Beside setting the `host` and `port`. The `useEmulator` function is still functioning the same.


Ehesp:
Hey thanks for this contribution.

The admin SDKs don't set the host/port directly like the client SDKs, instead they read from a environment variable, see https://firebase.google.com/docs/emulator-suite/connect_auth#admin_sdks

We should probably instead parse this value if it exists and pass it through.

rrousselGit:
>  I notice you has the "do_not_use_environment" lint rule set.

That's not the same "enrivonment". We're talking about using `Platform.environment`. Using `String.fromEnvironment` is completely unrelated.

Also, this would need testing :)

rrousselGit (changes requested):
Cf above

hn12404988:
Hi @rrousselGit , thanks for the feedback.
I tried to fix the point you mentioned and implemented the tests. Let me know if any issue.
Thanks for the review!

packages/dart_firebase_admin/lib/src/app/emulator.dart
```diff
@@ -0,0 +1,63 @@
...
+        port = 8080;
+
+  /// Try to get the Auth Emulator from the environment variable.
+  /// If not found, use the default Auth Emulator.
+  factory Emulator.auth() {
+    if (!Platform.environment.containsKey('FIREBASE_AUTH_EMULATOR_HOST')) {
+      return const Emulator._defaultAuth();
+    } else {
+      return Emulator.fromEnvString(
+        Platform.environment['FIREBASE_AUTH_EMULATOR_HOST']!,
```
rrousselGit, line 44: Could you remove this `!`?

This is possible by doing something like:

```dart
final env = Platform.environment['FIREBASE_AUTH_EMULATOR_HOST'];
if (env != null) {...}
else {...}
```

packages/dart_firebase_admin/lib/src/app/emulator.dart
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  /// Try to get the Firestore Emulator from the environment variable.
+  /// If not found, use the default Firestore Emulator.
+  factory Emulator.firestore() {
+    if (!Platform.environment.containsKey('FIRESTORE_EMULATOR_HOST')) {
+      return const Emulator._defaultFirestore();
+    } else {
+      return Emulator.fromEnvString(
+        Platform.environment['FIRESTORE_EMULATOR_HOST']!,
```
rrousselGit, line 56: Same here

packages/dart_firebase_admin/lib/src/app/emulator.dart
```diff
@@ -0,0 +1,63 @@
+part of '../app.dart';
+
+/// Represents a Firebase Emulator.
+/// Now for Auth and Firestore.
+class Emulator {
+
```
rrousselGit, line 6:
```suggestion
```

packages/dart_firebase_admin/lib/src/app/firebase_admin.dart
```diff
@@ -9,19 +9,46 @@ class FirebaseAdminApp {
...
+    Emulator? firestoreEmulator,
+  }) {
+    useAuthEmulator(emulator: authEmulator);
+    useFirestoreEmulator(emulator: firestoreEmulator);
+  }
+
+  /// Use the Firebase Auth Emulator to run the app locally.
+  void useAuthEmulator({
+    Emulator? emulator
+  }) {
```
rrousselGit, line 38: Wouldn't a simpler API be this?

```dart
void useAuthEmulator([Uri? uri])
```

We can then remove that `Emulator` class.

packages/dart_firebase_admin/lib/src/app/firebase_admin.dart
```diff
@@ -9,19 +9,46 @@ class FirebaseAdminApp {
...
+    _isUsingAuthEmulator = true;
+    emulator ??= Emulator.auth();
```
rrousselGit, line 40: I'd rather have the environment logic directly in here, instead of a constructor inside a separate Emulator class.

Issue #59, opened 2024-11-05
Transactions in Firestore


evandrobubiak:
It would be important to support transactions in Firestore. Many backend activities depend on this functionality to maintain data integrity.

BirjuVachhani:
+1

mafreud:
@rrousselGit It seems that @evandrobubiak has submitted a PR. Could you check it?

PR #62, opened 2024-11-18
Implements Transactions in Firestore

evandrobubiak:
- Implements Transactions for Firestore #59 
- Implements error handling for Firestore errors.
- Created some tests for Transactions.

### DIFF
--- a/packages/dart_firebase_admin/lib/src/google_cloud_firestore/firestore.dart
+++ b/packages/dart_firebase_admin/lib/src/google_cloud_firestore/firestore.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+import 'dart:convert';
 import 'dart:math' as math;
 
 import 'package:collection/collection.dart';
@@ -28,6 +30,7 @@ part 'write_batch.dart';
 part 'document_change.dart';
 part 'filter.dart';
 part 'firestore_exception.dart';
+part 'firestore_api_request_internal.dart';
 part 'collection_group.dart';
 
 class Firestore {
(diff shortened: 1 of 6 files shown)


packages/dart_firebase_admin/lib/src/google_cloud_firestore/firestore.dart
```diff
@@ -193,6 +196,79 @@ class Firestore {
...
+      return Future.error(
+        FirebaseFirestoreAdminException(
+          FirestoreClientErrorCode.failedPrecondition,
+          'Transaction max attemps must be > 0.',
+        ),
+      );
+    }
+    return Future<T>(() async {
+      var attempts = 1;
+      while (true) {
```
labrom, line 215: I think the following would be a more defensive style: 
```
var attempts = 0;
while(++attempts <= maxAttempts) {
```
and starting line 222:
```
if (attempts >= maxAttempts) {
  return Future.error(...);
}
```

packages/dart_firebase_admin/lib/src/google_cloud_firestore/firestore.dart
```diff
@@ -193,6 +196,79 @@ class Firestore {
...
+    return _client.v1(
+      (client) async {
+        final transactionResponse = await client.projects.databases.documents
+            .beginTransaction(transactionRequest, _formattedDatabaseName)
+            .catchError(_handleException);
+
+        final transactionId = transactionResponse.transaction;
+        final transaction = Transaction(this, transactionId!);
+        final result = await transactionHandler(transaction);
+        await transaction._transactionWriteBatch
```
labrom, line 264: Would it make sense to either have a getter for `transaction._transactionWriteBatch` or implement `commit()` and `catchError()` as part of `Transaction`? This would avoid exposing a private member (`_transactionWriteBatch`).

rrousselGit (changes requested):
I appreciate, but for the sake of maintenance we're trying to stick to a port of the official NodeJS implementation as much as possible.

Cf https://github.com/googleapis/nodejs-firestore/blob/07a88420062c428b04d61638c6ba432422e01652/dev/src/transaction.ts#L79
https://github.com/googleapis/nodejs-firestore/blob/07a88420062c428b04d61638c6ba432422e01652/dev/src/transaction.ts#L582

Do you mind updating your PR such that it is a port of the original code?

rrousselGit:
Thanks! I understand that this is quite a big change.

evandrobubiak:
Close this requeset and will open a new one with all new code based on NodeJS official implementation.

PR #81, opened 2025-04-13
Interruption Level for APS to deliver time-sensitive notifications for IOS

### LINKED ISSUE #80: Cant make Aps Time Sensitive
Just add the field, Why is _toProto private? Why is it impossible to add custom properties if not all the properties were filled out? These change from update to update, i'd recommend not restricting stuff for no reason just because it feels cleaner.

cyberpwnn:
fixes #80

Simply adds time-sensitive as a property for APS and allows it to play instead of getting rejected on firebases end.

You can now deliver time sensitive notifications with sound on IOS 

* Added String? interruptionLevel and String? soundRaw to the Aps model
* If sound is defined, that is serialized, otherwise soundRaw is, if neither are defined, the sound field is not included in the map.

```dart
MulticastMessage(
  tokens: ...,
  data: ...,
  notification: Notification(title: title, body: body),
  apns: ApnsConfig(
    headers: {"apns-priority": "5"}, // Firebase requires 5, anything higher is rejected
    payload: ApnsPayload(
      aps: Aps(
        soundRaw: "default", // Makes it play a sound on ios. Using "CriticalSound" object rejects the notification on firebase.
        interruptionLevel: "time-sensitive", // Actually makes the notification time sensitive
      ),
    ),
  ),
)
```

This is the correct way to deliver Aps payload with time-sensitive and with actual sound, the CriticalSound object is not up to aps spec and it just gets rejected by firebase, notifications wont even deliver using that.

### DIFF
--- a/packages/dart_firebase_admin/lib/src/messaging/messaging_api.dart
+++ b/packages/dart_firebase_admin/lib/src/messaging/messaging_api.dart
@@ -452,6 +452,7 @@ class Aps {
     this.mutableContent,
     this.category,
     this.threadId,
+    this.interruptionLevel,
   });
 
   /// Alert to be included in the message. This may be a string or an object of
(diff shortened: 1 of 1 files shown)


demolaf:
Hi @cyberpwnn thanks for the PR! This feature would be added in this PR https://github.com/firebase/firebase-admin-dart/pull/219 , I'll be closing this now.

Feel free to open a new PR if the issue still persists.

PR #84, opened 2025-04-28
Fix #83: Removed unnecessary check

### LINKED ISSUE #83: Firestore map keys considered invalid when containing `/` characters
**Describe the bug**
We often store mappings of document reference paths (as strings) to values inside documents.

e.g.
```
{
    'activityType': 'activityA',
    'agents': [
      {'products/existing-product': 5.0},
    ],
}
```

Using the native Firestore libraries, we encountered no issues writing Firestore maps inside our documents that contain `/` characters in the key string. This seemingly is allowed in Firestore documents but triggers the `Invalid argument (path)` exception when attempting to write a document containing the map using the library.

```
Invalid argument (path): Paths can't be empty and must not contain '*~/[]'.: "products/existing-product"
```
**To Reproduce**

```dart
   await firestore.doc("activities/new-activity").set({
        'activityType': 'activityA',
        'agents': [
          {'products/product-a': 5.0},
        ],
      });
```

**Expected behavior**
This should be a valid document and written to Firestore with the string key `" products/product-a"` in the document's map list.

HeySreelal:
## Related Issues

Fixes #83

---

Changes:
1. Removed the unnecessary check in the `_StringFieldMask` constructor that thrown `ArgumentError` when field paths contained any of `*~/[]` characters. All of these characters are allowed in Firestore field names it seems.

### DIFF
--- a/packages/dart_firebase_admin/lib/src/google_cloud_firestore/path.dart
+++ b/packages/dart_firebase_admin/lib/src/google_cloud_firestore/path.dart
@@ -231,8 +231,6 @@ sealed class FieldMask {
   factory FieldMask.fieldPath(FieldPath fieldPath) = _FieldPathFieldMask;
 }
 
-final _fieldPathRegex = RegExp(r'^[^*~/[\]]+$');
-
 class _StringFieldMask implements FieldMask {
   _StringFieldMask(this.path) {
     if (path.contains('..')) {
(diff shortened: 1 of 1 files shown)


rrousselGit:
I appreciate, but that's not the right fix. The error shouldn't be removed. It's present in the JS SDK.
It's about making sure that the error doesn't trigger in cases where it shouldn't

HeySreelal:
Totally understand.. I missed to check the JS SDK first. 

Removed the check because the `_fieldPathRegex` was only used in the `_StringFieldMask` constructor. Since this is a fork from JS SDK, it's best to close this. Sorry though :)

PR #85, opened 2025-05-28
expose the payload on a decoded token to allow access to custom claims

markbreuss:
If custom claims have been set via a custom token creation they are currently not accessible via the decoded token.
This PR exposes the original payload to access custom claims.
https://firebase.google.com/docs/auth/admin/create-custom-tokens#database-rules_1

It also finishes the todo in the code.

Let me know what you think.


rrousselGit (changes requested):
Could you add tests for this?

demolaf:
Hi @markbreuss , thanks for opening this PR. Could you please pull the most recent changes and add tests for this as well?

PR #211, opened 2026-04-09
doc(dart_firebase_admin): s/undefined/null

brianquinlan:
Removes the use of `undefined` in dartdoc (`undefined` is not a type in Dart).


Ehesp:
Thats the javascripter coming through!

demolaf (approved):

PR #216, opened 2026-04-10
fix: oops - dependency override removal

kevmoo:


demolaf (approved):
@kevmoo thanks for catching this, I thought I got all of them while testing 🥲.

PR #221, opened 2026-04-14
chore: update version constraints for examples

kevmoo:


lahirumaramba (approved):

packages/firebase_admin_sdk/example/pubspec.yaml
```diff
@@ -7,6 +7,6 @@ environment:
   sdk: '^3.9.0'
 
 dependencies:
-  firebase_admin_sdk:
-  google_cloud_firestore:
+  firebase_admin_sdk: ^0.5.0
+  google_cloud_firestore: ^0.5.0
```
lahirumaramba, line 11: just a note: are these apps part of CIs? If they are the tests will fail till the package version is published

lahirumaramba (approved):

packages/firebase_admin_sdk/example/pubspec.yaml
kevmoo: because they are IN this repo and part of the workspace, they are fine.

kevmoo: It's important for the examples to be "honest" about the version they support because folks will copy-paste.

PR #231, opened 2026-04-15
chore: coverage scripting fixes

kevmoo:


scripts/process_coverage.dart
```diff
@@ -0,0 +1,131 @@
...
+    return (0.00, 0, 0);
+  }
+
+  final adminCov = calculateCoverage(savedAdminFile);
+  final firestoreCov = calculateCoverage(savedFirestoreFile);
+
+  // Calculate total coverage from merged file
+  final totalCov = calculateCoverage(File('$adminPkgDir/coverage.lcov'));
+
+  final coveragePct = totalCov.$1.toStringAsFixed(2);
```
brianquinlan: I think that you can use the record names here, right?

scripts/process_coverage.dart
```diff
@@ -0,0 +1,131 @@
...
+  final repoRoot = Directory(scriptDir).parent.path;
+
+  final adminPkgDir = '$repoRoot/packages/firebase_admin_sdk';
+  final firestorePkgDir = '$repoRoot/packages/google_cloud_firestore';
+
+  final adminCoverageFile = File('$adminPkgDir/coverage.lcov');
+  final firestoreCoverageFile = File('$firestorePkgDir/coverage.lcov');
+
+  // 1. Save individual package coverage files before merging
+  final savedAdminFile = File('$adminPkgDir/coverage_admin.lcov');
```
brianquinlan: Why do you need to do this? As far as I can tell, you aren't mutating `adminCoverageFile` or `firestoreCoverageFile`.

scripts/process_coverage.dart
```diff
@@ -0,0 +1,131 @@
...
+  if (adminCoverageFile.existsSync()) {
+    adminCoverageFile.copySync(savedAdminFile.path);
+  }
+  if (firestoreCoverageFile.existsSync()) {
+    firestoreCoverageFile.copySync(savedFirestoreFile.path);
+  }
+
+  // 2. Merge coverage reports
+  final coverageFiles = <File>[];
+  if (adminCoverageFile.existsSync()) coverageFiles.add(adminCoverageFile);
```
brianquinlan: These files should always be present, right?

scripts/process_coverage.dart
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  if (coverageFiles.isEmpty) {
+    print('No coverage files found!');
+    exitCode = 1;
+    return;
+  }
+
+  // Create merged file content
+  final mergedContent = StringBuffer();
```
brianquinlan, line 51: Is there a reason why you are writing to a `StringBuffer` and then writing the entire buffer to a file?

Wouldn't this be more direct:
```dart
  final mergedFile = File('$adminPkgDir/coverage.lcov')..openWrite();
  for (final file in coverageFiles) {
    mergedFile.writeln(file.readAsStringSync());
  }
  await mergedFile.close();
```

scripts/process_coverage.dart
```diff
@@ -0,0 +1,131 @@
...
+    mergedContent.writeln(file.readAsStringSync());
+  }
+
+  // Overwrite coverage.lcov in admin package with merged content
+  File(
+    '$adminPkgDir/coverage.lcov',
+  ).writeAsStringSync(mergedContent.toString());
+
+  // 3. Calculate coverage and check threshold
+  (double pct, int hit, int total) calculateCoverage(File file) {
```
brianquinlan: Is there a reason why this function must be nested?

scripts/process_coverage.dart
```diff
@@ -0,0 +1,131 @@
...
+
+  // Calculate total coverage from merged file
+  final totalCov = calculateCoverage(File('$adminPkgDir/coverage.lcov'));
+
+  final coveragePct = totalCov.$1.toStringAsFixed(2);
+  final totalLines = totalCov.$3;
+  final hitLines = totalCov.$2;
+
+  // Output for GitHub Actions
+  void githubOutput(String key, String value) {
```
brianquinlan: Does this have to be a nested function? I'm also not a big fan of the variable and function name being the same.

scripts/process_coverage.dart
```diff
@@ -0,0 +1,131 @@
...
+  githubOutput('total_lines', totalLines.toString());
+  githubOutput('hit_lines', hitLines.toString());
+
+  githubOutput('admin_coverage', adminCov.$1.toStringAsFixed(2));
+  githubOutput('firestore_coverage', firestoreCov.$1.toStringAsFixed(2));
+
+  // Console output
+  print('=== Coverage Report ===');
+  print(
+    'firebase_admin_sdk: ${adminCov.$1.toStringAsFixed(2)}% (${adminCov.$2}/${adminCov.$3} lines)',
```
brianquinlan: In some places this script is generic (i.e. working in terms of `coverageFiles`), while in other cases, like this, you require exact files.

scripts/process_coverage.dart
kevmoo: Meh.

kevmoo:
@brianquinlan - this was a robot rewrite of the existing bash script – warts and all.

PTAL

brianquinlan (approved):

PR #233, opened 2026-04-15
fix: drop googleapis_beta dependency

kevmoo:
Trivial to inline!

### DIFF
--- a/packages/firebase_admin_sdk/lib/src/app_check/app_check_http_client.dart
+++ b/packages/firebase_admin_sdk/lib/src/app_check/app_check_http_client.dart
@@ -72,15 +72,6 @@ class AppCheckHttpClient {
     (client, projectId) => fn(appcheck1.FirebaseappcheckApi(client), projectId),
   );
 
-  /// Executes an App Check v1Beta API operation with automatic projectId injection.
-  Future<R> v1Beta<R>(
-    Future<R> Function(appcheck1_beta.FirebaseappcheckApi api, String projectId)
-    fn,
-  ) => _run(
-    (client, projectId) =>
-        fn(appcheck1_beta.FirebaseappcheckApi(client), projectId),
-  );
-
   /// Exchange a custom token for an App Check token (low-level API call).
   ///
   /// Returns the raw googleapis response without transformation.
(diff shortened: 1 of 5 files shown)


kevmoo:
Let's drop a package that hasn't been updated in 2+ years!

packages/firebase_admin_sdk/lib/src/app_check/app_check_http_client.dart
```diff
@@ -101,15 +92,34 @@ class AppCheckHttpClient {
...
+  if (response.statusCode != 200) {
+    throw appcheck1.DetailedApiRequestError(
+      response.statusCode,
+      response.body,
+      jsonResponse: jsonDecode(response.body) as Map<String, dynamic>?,
+    );
   }
+
+  final data = jsonDecode(response.body) as Map<String, dynamic>;
+  return data['alreadyConsumed'] as bool? ?? false;
```
brianquinlan: Isn't this the same as:

`return data['alreadyConsumed'] == true;`

brianquinlan (approved):

packages/firebase_admin_sdk/lib/src/app_check/app_check_http_client.dart
```diff
@@ -101,15 +92,34 @@ class AppCheckHttpClient {
...
+    Uri.parse(url),
+    headers: {'Content-Type': 'application/json', 'Accept': 'application/json'},
+    body: jsonEncode({'appCheckToken': token}),
+  );
+
+  if (response.statusCode != 200) {
+    throw appcheck1.DetailedApiRequestError(
+      response.statusCode,
+      response.body,
+      jsonResponse: jsonDecode(response.body) as Map<String, dynamic>?,
```
brianquinlan: I'd at least handle the case where the response is non-JSON if the result is not 200.

packages/firebase_admin_sdk/lib/src/app_check/app_check_http_client.dart
kevmoo: Yep. Just the auto-generated bits being silly.

packages/firebase_admin_sdk/lib/src/app_check/app_check_http_client.dart
kevmoo: ack. Again, this is from auto-generated APIs.

PR #244, opened 2026-04-16
chore: group dependabot so they are clustered into one PR

kevmoo:


kevmoo:
Less noise next month!

demolaf (approved):

PR #248, opened 2026-04-21
chore: switch to using google_cloud_firestore_v1

brianquinlan:
Update `package:google_cloud_firestore` to use `package:google_cloud_firestore_v1` in place of `package:googleapis`. The indention is that the public API should not have changed.

Sorry, this is going to be a nightmare to review. I tried to keep the incidental changes to a minimum.

There are no changes to the integration tests so it is at least possible that this change is correct and didn't break the public API ;-)

### DIFF
--- a/packages/google_cloud_firestore/lib/src/serializer.dart
+++ b/packages/google_cloud_firestore/lib/src/serializer.dart
@@ -100,6 +100,9 @@ class Serializer {
       case _Serializable():
         return value._toProto();
 
+      case Uint8List():
+        return firestore_v1.Value(bytesValue: value);
+
       case List():
         return firestore_v1.Value(
           arrayValue: firestore_v1.ArrayValue(
@@ -118,10 +121,6 @@ class Serializer {
         if (fields.fields.isEmpty) return null;
 
         return firestore_v1.Value(mapValue: fields);
-
-      case Uint8List():
-        return firestore_v1.Value(bytesValue: value);
-
       default:
         throw ArgumentError.value(
           value,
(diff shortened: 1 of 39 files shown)


demolaf:
> Update `package:google_cloud_storage` to use `package:google_cloud_firestore_v1` in place of `package:googleapis`. The indention is that the public API should not have changed.
> 
> Sorry, this is going to be a nightmare to review. I tried to keep the incidental changes to a minimum.
> 
> There are no changes to the integration tests so it is at least possible that this change is correct and didn't break the public API ;-)

Currently reviewing this 😅! I think you meant `package:google_cloud_firestore` in the PR description?

brianquinlan:
> > Update `package:google_cloud_storage` to use `package:google_cloud_firestore_v1` in place of `package:googleapis`. The indention is that the public API should not have changed.
> > Sorry, this is going to be a nightmare to review. I tried to keep the incidental changes to a minimum.
> > There are no changes to the integration tests so it is at least possible that this change is correct and didn't break the public API ;-)
> 
> Currently reviewing this 😅! I think you meant `package:google_cloud_firestore` in the PR description?

Yes, fixed.

packages/google_cloud_firestore/lib/src/serializer.dart
demolaf, line 1: I see, since this new sdk i.e. google_cloud_firestore_v1 uses protobuf which supports int64 natively and infinity, -infinity and Nan unlike JSON, so no need for the workaround 👍.

demolaf (approved):
LGTM

demolaf:
@brianquinlan the PR branch needs to be updated with `main`.

PR #249, opened 2026-04-26
feat(remote_config): add Firebase Remote Config support

long1eu:
## Summary

Adds Firebase Remote Config support to the Dart Admin SDK: template management (read, validate, publish with force, rollback, list versions, parse from JSON) and server-side template evaluation (fetch + in-process rules engine for percent rollouts and string/numeric/semver custom signals).

## Notable design choices

- **Direct REST calls** instead of `googleapis`-generated client — the generated client doesn't expose response headers (no ETag) and doesn't accept `If-Match`, both of which are required for optimistic concurrency and force-publish.
- **Server-template publishing is not exposed** — the REST API doesn't accept writes on the `firebase-server` namespace; server templates are Console-managed only.
- Field shapes and required/output-only annotations match the v1 discovery schema.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -287,6 +287,15 @@ jobs:
       - name: Run WIF auth test
         run: dart test test/integration/app/firebase_app_prod_test.dart --concurrency=1 --tags wif
 
+      - name: Run Remote Config integration tests
+        env:
+          SERVICE_ACCOUNT: ${{ secrets.SERVICE_ACCOUNT }}
+        run: |
+          project_id="${SERVICE_ACCOUNT##*@}"
+          project_id="${project_id%%.iam.gserviceaccount.com}"
+          RC_TEST_PROJECT_ID="$project_id" \
+            dart test test/integration/remote_config/ --concurrency=1
+
   publish:
     name: Publish verification
     runs-on: ubuntu-latest
--- a/packages/firebase_admin_sdk/CHANGELOG.md
+++ b/packages/firebase_admin_sdk/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 0.5.2-wip
 
+- Add Remote Config support: template management and server-side template evaluation.
 - Remove dependency on `package:equatable`.
 - Make `Query`, `CollectionReference`, `DocumentReference`, and `CollectionGroup` mockable.
 - `Credential.createClient(List<String> scopes)` — create an authenticated `AuthClient` directly
--- a/packages/firebase_admin_sdk/lib/src/remote_config/condition_evaluator.dart
+++ b/packages/firebase_admin_sdk/lib/src/remote_config/condition_evaluator.dart
@@ -1,0 +1,281 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+part of 'remote_config.dart';
+
+/// Server-side condition evaluator for [ServerTemplate.evaluate].
+@internal
+class ConditionEvaluator {
+  static const int _maxRecursionDepth = 10;
+
+  /// Evaluates each named condition in [namedConditions] against [context],
+  /// returning a map from condition name to boolean result. The returned map
+  /// preserves insertion order, matching the priority order of the input.
+  Map<String, bool> evaluateConditions(
+    List<NamedCondition> namedConditions,
+    EvaluationContext context,
+  ) {
+    final results = <String, bool>{};
+    for (final namedCondition in namedConditions) {
+      results[namedCondition.name] = _evaluate(
+        namedCondition.condition,
+        context,
+        0,
+      );
+    }
+    return results;
+  }
+
+  bool _evaluate(
+    OneOfCondition condition,
+    EvaluationContext context,
+    int nestingLevel,
+  ) {
+    if (nestingLevel >= _maxRecursionDepth) return false;
+    return switch (condition) {
+      TrueCondition() => true,
+      FalseCondition() => false,
+      OrCondition() => _evaluateOr(condition, context, nestingLevel + 1),
+      AndCondition() => _evaluateAnd(condition, context, nestingLevel + 1),
+      PercentCondition() => _evaluatePercent(condition, context),
+      CustomSignalCondition() => _evaluateCustomSignal(condition, context),
+    };
+  }
+
+  bool _evaluateOr(
+    OrCondition cond,
+    EvaluationContext context,
+    int nestingLevel,
+  ) {
+    final subs = cond.conditions ?? const <OneOfCondition>[];
+    for (final sub in subs) {
+      if (_evaluate(sub, context, nestingLevel + 1)) return true;
+    }
+    return false;
+  }
+
+  bool _evaluateAnd(
+    AndCondition cond,
+    EvaluationContext context,
+    int nestingLevel,
+  ) {
+    final subs = cond.conditions ?? const <OneOfCondition>[];
+    for (final sub in subs) {
+      if (!_evaluate(sub, context, nestingLevel + 1)) return false;
+    }
+    return true;
+  }
+
+  bool _evaluatePercent(PercentCondition cond, EvaluationContext context) {
+    final randomizationId = context.randomizationId;
+    if (randomizationId == null) return false;
+
+    final op = cond.percentOperator;
+    if (op == null || op == PercentConditionOperator.unknown) return false;
+
+    final microPercent = cond.microPercent ?? 0;
+    final upper = cond.microPercentRange?.microPercentUpperBound ?? 0;
+    final lower = cond.microPercentRange?.microPercentLowerBound ?? 0;
+
+    final seed = cond.seed;
+    final seedPrefix = (seed != null && seed.isNotEmpty) ? '$seed.' : '';
+    final stringToHash = '$seedPrefix$randomizationId';
+
+    final hash = _hashSeededRandomizationId(stringToHash);
+    final instanceMicroPercentile = (hash % BigInt.from(100 * 1000000)).toInt();
+
+    return switch (op) {
+      PercentConditionOperator.lessOrEqual =>
+        instanceMicroPercentile <= microPercent,
+      PercentConditionOperator.greaterThan =>
+        instanceMicroPercentile > microPercent,
+      PercentConditionOperator.between =>
+        instanceMicroPercentile > lower && instanceMicroPercentile <= upper,
+      PercentConditionOperator.unknown => false,
+    };
+  }
+
+  /// SHA-256 the input string and return the resulting digest as a [BigInt].
+  /// Visible for testing.
+  @internal
+  static BigInt hashSeededRandomizationIdForTest(
+    String seededRandomizationId,
+  ) => _hashSeededRandomizationId(seededRandomizationId);
+
+  static BigInt _hashSeededRandomizationId(String input) {
+    final bytes = sha256.convert(utf8.encode(input)).bytes;
+    final hex = StringBuffer();
+    for (final b in bytes) {
+      hex.write(b.toRadixString(16).padLeft(2, '0'));
+    }
+    return BigInt.parse(hex.toString(), radix: 16);
+  }
+
+  bool _evaluateCustomSignal(
+    CustomSignalCondition cond,
+    EvaluationContext context,
+  ) {
+    final op = cond.customSignalOperator;
+    final key = cond.customSignalKey;
+    final targets = cond.targetCustomSignalValues;
+
+    if (op == null ||
+        op == CustomSignalOperator.unknown ||
+        key == null ||
+        targets == null ||
+        targets.isEmpty) {
+      return false;
+    }
+
+    final actual = context.customSignals[key];
+    if (actual == null) return false;
+
+    return switch (op) {
+      CustomSignalOperator.stringContains => _compareStrings(
+        targets,
+        actual,
+        (target, actualString) => actualString.contains(target),
+      ),
+      CustomSignalOperator.stringDoesNotContain => !_compareStrings(
+        targets,
+        actual,
+        (target, actualString) => actualString.contains(target),
+      ),
+      CustomSignalOperator.stringExactlyMatches => _compareStrings(
+        targets,
+        actual,
+        (target, actualString) => actualString.trim() == target.trim(),
+      ),
+      CustomSignalOperator.stringContainsRegex => _compareStrings(
+        targets,
+        actual,
+        (target, actualString) => RegExp(target).hasMatch(actualString),
+      ),
+      CustomSignalOperator.numericLessThan => _compareNumbers(
+        actual,
+        targets[0],
+        (r) => r < 0,
+      ),
+      CustomSignalOperator.numericLessEqual => _compareNumbers(
+        actual,
+        targets[0],
+        (r) => r <= 0,
+      ),
+      CustomSignalOperator.numericEqual => _compareNumbers(
+        actual,
+        targets[0],
+        (r) => r == 0,
+      ),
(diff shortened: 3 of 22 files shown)


long1eu:
@kevmoon @Lyokone can you have a look? :D

demolaf:
Hi @long1eu , thanks for the PR! 

At the moment there's no support for Remote Config in the `googleapis` generated package and all our Firebase services are routed through it.

I'll get some more details on what the plan is there and keep this PR up to date.

long1eu:
Thanks for taking a look @demolaf!

Quick context that might be useful while you check on the `googleapis` plan:

- `firebaseremoteconfig` *was* in `package:googleapis` up through `13.2.0`; it was dropped between `13.2.0` and `14.0.0` and isn't in `16.0.0` (the version this repo currently depends on). So even if I had wanted to route through it, the generated client isn't shipped today.

- More importantly, four of the five admin endpoints (`getTemplate`, `publishTemplate`, `validateTemplate`, `rollback`) require reading and setting the `ETag` / `If-Match` headers — that's how Remote Config does optimistic concurrency, and how `force=true` is expressed (`If-Match: *`). The googleapis-generated clients don't surface request/response headers, so even when `firebaseremoteconfig` was available in `googleapis`, those four endpoints would still need a raw-REST path. Only `listVersions` is a clean GET that could go through the generated client.

Happy to take this in any direction that fits the project's plans:

1. **Leave as-is** — raw REST throughout, consistent within the new `remote_config` library.
2. **Hybrid** — route `listVersions` through `googleapis` if/when `firebaseremoteconfig` is re-added; keep raw REST for the four ETag-bound endpoints.
3. **Wait** for the broader `googleapis`-package decision before merging, and refactor accordingly.

Just let me know which you'd prefer once you have more details on the plan.

Ortes:
Production user data point, in case it helps this get unblocked.

We run Dart Cloud Functions in production alongside a legacy Node codebase, and migrate `onCall` endpoints to Dart as capabilities land. Missing server-side Remote Config is the one capability gap blocking our next migration: the endpoint reads a kill switch through

```js
const template = await getRemoteConfig().getServerTemplate({defaultConfig: {[key]: false}});
return template.evaluate().getBoolean(key);
```

which maps 1:1 onto `RemoteConfig.getServerTemplate` → `ServerTemplate.evaluate` → `ServerConfig.getBoolean` in this PR. I read through the branch; the shape follows the Node SDK closely enough that it would slot in behind our existing helper without touching call sites.

On the `googleapis` question from @demolaf: `firebaseremoteconfig` isn't in the generated set today — zero occurrences in `config.yaml` on `google/googleapis.dart`, across the stable, beta and `skipped_apis` lists alike. The changelog puts its removal in `14.0.0`, which lines up with what @long1eu described, so there's nothing to route through as things stand. His `ETag` / `If-Match` point also reads as correct to me for `getTemplate` / `publishTemplate` / `validateTemplate` / `rollback`: those need request *and* response headers that the generated clients don't surface, which would leave only `listVersions` as a candidate even if the API were regenerated.

Would a maintainer be able to pick between the three options @long1eu offered in April? Even "option 3, wait for the googleapis decision" would help — the PR has gone `CONFLICTING`, and nobody can tell whether rebasing it is worth the contributor's time. `maintainerCanModify` is true on it if that makes things easier.

Also worth noting there was no issue tracking Remote Config in this repo, so there's been nothing for users to find or subscribe to. I've opened #323 so the capability is trackable separately from the PR.

Happy to test a rebased branch against a real Firebase project (server template with percent conditions and custom signals) and report back if that's useful.

demolaf:
Hi @long1eu , can you rebase this on the latest version on `main`?

packages/firebase_admin_sdk/test/integration/remote_config/remote_config_test.dart
```diff
@@ -0,0 +1,319 @@
...
+  valueType: ParameterValueType.string,
+);
+
+void main() {
+  // Three conditions must hold for the suite to run:
+  //   - RC_TEST_PROJECT_ID is set (target project for the round-trip).
+  //   - Credentials are available (CI WIF sets GOOGLE_APPLICATION_CREDENTIALS;
+  //     local opt-in via RUN_PROD_TESTS=true).
+  final projectId = _rcProjectId;
+  final shouldRun = projectId != null && (hasWifEnv || hasProdEnv);
```
demolaf, line 69: Can we tag this group `prod` and drop the `shouldRun` and `skipReason` guard? `hasWifEnv` and `hasProdEnv` were valid when this was written but #255 removed them two days later, and `dart_test.yaml` now excludes `prod` by default, which is what the guard was doing by hand.

packages/firebase_admin_sdk/lib/src/remote_config/remote_config_exception.dart
```diff
@@ -0,0 +1,147 @@
...
+
+String? _serverErrorMessage(Object? response) {
+  if (response is Map) {
+    final error = response['error'];
+    if (error is Map && error['message'] is String) {
+      return error['message'] as String;
+    }
+  }
+  return null;
+}
```
demolaf, line 127: Can we collapse these into map patterns? It removes both `as` casts and the `is` guards, and matches how the same extraction reads in `firestore_exception.dart`.

```suggestion
String? _serverErrorCode(Object? response) => switch (response) {
  {'error': final String code} => code,
  {'error': {'status': final String status}} => status,
  {'error': {'code': final String code}} => code,
  _ => null,
};

String? _serverErrorMessage(Object? response) => switch (response) {
  {'error': {'message': final String message}} => message,
  _ => null,
};
```

packages/firebase_admin_sdk/lib/src/remote_config/condition_evaluator.dart
```diff
@@ -0,0 +1,281 @@
...
+  ) => _hashSeededRandomizationId(seededRandomizationId);
+
+  static BigInt _hashSeededRandomizationId(String input) {
+    final bytes = sha256.convert(utf8.encode(input)).bytes;
+    final hex = StringBuffer();
+    for (final b in bytes) {
+      hex.write(b.toRadixString(16).padLeft(2, '0'));
+    }
+    return BigInt.parse(hex.toString(), radix: 16);
+  }
```
demolaf, line 123: Fold the bytes directly rather than building 64 hex chars and re-parsing them. Same value, one expression.

```suggestion
  static BigInt _hashSeededRandomizationId(String input) {
    return sha256
        .convert(utf8.encode(input))
        .bytes
        .fold(BigInt.zero, (acc, b) => (acc << 8) | BigInt.from(b));
  }
```

packages/firebase_admin_sdk/lib/src/remote_config/remote_config_request_handler.dart
```diff
@@ -0,0 +1,163 @@
...
+      throw FirebaseRemoteConfigException(
+        RemoteConfigErrorCode.invalidArgument,
+        'ETag header missing from response.',
+      );
+    }
+    return RemoteConfigTemplate.fromJson(<String, Object?>{
+      ...result.body,
+      'etag': result.etag,
+    });
+  }
```
demolaf, line 121: Can we extract the etag check? It is duplicated at :49 and :98, and each copy needs a `!` right after proving non-null.

```suggestion
  String _requireEtag(String? etag, [String source = 'response']) =>
      switch (etag) {
        final e? when e.isNotEmpty => e,
        _ => throw FirebaseRemoteConfigException(
          RemoteConfigErrorCode.invalidArgument,
          'ETag header missing from $source.',
        ),
      };

  RemoteConfigTemplate _parseTemplate(RemoteConfigHttpResult result) {
    return RemoteConfigTemplate.fromJson(<String, Object?>{
      ...result.body,
      'etag': _requireEtag(result.etag),
    });
  }
```

packages/firebase_admin_sdk/lib/src/remote_config/remote_config_request_handler.dart
```diff
@@ -0,0 +1,163 @@
...
+      throw FirebaseRemoteConfigException(
+        RemoteConfigErrorCode.invalidArgument,
+        'ETag header missing from validateTemplate response.',
+      );
+    }
+    final parsed = RemoteConfigTemplate.fromJson(<String, Object?>{
+      ...result.body,
+      'etag': template.etag,
+    });
+    return parsed;
```
demolaf, line 59: Use the shared `_requireEtag` here. This also drops the `parsed` local, which is only assigned and returned.

```suggestion
    _requireEtag(result.etag, 'validateTemplate response');
    return RemoteConfigTemplate.fromJson(<String, Object?>{
      ...result.body,
      'etag': template.etag,
    });
```

packages/firebase_admin_sdk/lib/src/remote_config/remote_config_request_handler.dart
```diff
@@ -0,0 +1,163 @@
...
+    if (result.etag == null || result.etag!.isEmpty) {
+      throw FirebaseRemoteConfigException(
+        RemoteConfigErrorCode.invalidArgument,
+        'ETag header missing from getServerTemplate response.',
+      );
+    }
+    return ServerTemplateData.fromJson(<String, Object?>{
+      ...result.body,
+      'etag': result.etag,
+    });
```
demolaf, line 107: Same here, folding the check into the value it guards.

```suggestion
    return ServerTemplateData.fromJson(<String, Object?>{
      ...result.body,
      'etag': _requireEtag(result.etag, 'getServerTemplate response'),
    });
```

packages/firebase_admin_sdk/lib/src/remote_config/remote_config.dart
```diff
@@ -0,0 +1,302 @@
...
+    if (json.isEmpty) {
+      throw FirebaseRemoteConfigException(
+        RemoteConfigErrorCode.invalidArgument,
+        'JSON string must be a non-empty string.',
+      );
+    }
+    Object? decoded;
+    try {
+      decoded = jsonDecode(json);
+    } catch (e) {
```
demolaf, line 107: Narrow this to the exception `jsonDecode` actually throws, so a programmer error here is not silently relabelled as `invalid-argument`.

```suggestion
    } on FormatException catch (e) {
```

packages/firebase_admin_sdk/test/integration/remote_config/remote_config_test.dart
```diff
@@ -0,0 +1,319 @@
...
+        //    succeeds despite the stale ETag in the candidate.
+        final published = await rc.publishTemplate(staleCandidate, force: true);
+        expect(
+          published.parameters.containsKey('dart_admin_sdk_e2e_marker'),
+          isTrue,
+        );
+
+        // 4. Cleanup: roll back to baseline so subsequent tests see the
+        //    original state.
+        final priorVersion = baseline.version?.versionNumber;
```
demolaf, line 269: Move this rollback into an `addTearDown` registered before the publish, and do the same at :185. As written it is skipped whenever an earlier assertion fails, which leaves the live project on the force-published template.

packages/firebase_admin_sdk/lib/src/remote_config/remote_config.dart
```diff
@@ -0,0 +1,302 @@
...
+        );
+      }
+    }
+
+    return ServerConfig.internal(configValues);
+  }
+
+  /// Returns the cached [ServerTemplateData], or throws if no template has
+  /// been loaded.
+  ServerTemplateData toJson() {
```
demolaf, line 291: Can we add a `toJson()` to `ServerTemplateData`? `jsonEncode` on the value returned here throws, so there is no way to persist a fetched server template even though `set()` accepts a JSON string.

packages/firebase_admin_sdk/lib/src/remote_config/remote_config.dart
```diff
@@ -0,0 +1,302 @@
...
+        'Failed to parse the JSON string: $json. $e',
+      );
+    }
+    if (decoded is! Map<String, Object?>) {
+      throw FirebaseRemoteConfigException(
+        RemoteConfigErrorCode.invalidArgument,
+        'JSON must decode to an object, got ${decoded.runtimeType}.',
+      );
+    }
+    return RemoteConfigTemplate.fromJson(decoded);
```
demolaf, line 119: Wrap this `fromJson` call so non-Firebase errors rethrow as `invalid-argument`. JSON that parses but has the wrong shape currently escapes as a raw `TypeError`, which callers following the documented contract will not catch. `ServerTemplate.set(String)` needs the same treatment.

packages/firebase_admin_sdk/lib/src/remote_config/remote_config_api.dart
```diff
@@ -0,0 +1,1350 @@
...
+    List<String>? targetCustomSignalValues,
+  }) : targetCustomSignalValues = targetCustomSignalValues == null
+           ? null
+           : List<String>.unmodifiable(targetCustomSignalValues),
+       super._() {
+    if (targetCustomSignalValues != null) {
+      if (targetCustomSignalValues.isEmpty) {
+        throw FirebaseRemoteConfigException(
+          RemoteConfigErrorCode.invalidArgument,
+          'targetCustomSignalValues must contain at least one value.',
```
demolaf, line 880: Can the `fromJson` paths skip these bounds checks and let the evaluator return false instead? An out-of-spec value from the server currently aborts the whole template parse, unlike `OneOfCondition.fromJson` which decodes unknown shapes to false.

packages/firebase_admin_sdk/lib/src/remote_config/remote_config_api.dart
```diff
@@ -0,0 +1,1350 @@
...
+          : Map<String, Object>.unmodifiable(defaultConfig);
+
+  /// Default config values used by [ServerConfig] for keys not defined in the
+  /// evaluated template. Values must be `String`, `num`, or `bool`.
+  final Map<String, Object>? defaultConfig;
+}
+
+/// Options for [RemoteConfig.initServerTemplate].
+class InitServerTemplateOptions extends GetServerTemplateOptions {
+  InitServerTemplateOptions({super.defaultConfig, this.template});
```
demolaf, line 1345: Either wire `GetServerTemplateOptions` and `InitServerTemplateOptions` into the method signatures or drop them before release. They are exported but unreferenced, and they read as the intended call shape for anyone porting from the Node SDK.

packages/firebase_admin_sdk/lib/src/remote_config/condition_evaluator.dart
```diff
@@ -0,0 +1,281 @@
...
+      ),
+      CustomSignalOperator.stringExactlyMatches => _compareStrings(
+        targets,
+        actual,
+        (target, actualString) => actualString.trim() == target.trim(),
+      ),
+      CustomSignalOperator.stringContainsRegex => _compareStrings(
+        targets,
+        actual,
+        (target, actualString) => RegExp(target).hasMatch(actualString),
```
demolaf, line 163: Catch `FormatException` around this `RegExp` and treat an uncompilable pattern as a non-match. A malformed Console-authored target currently throws out of the synchronous `evaluate()` and keeps throwing for the life of the cached template.

packages/firebase_admin_sdk/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## 0.5.2-wip
 
+- Add Remote Config support: template management and server-side template evaluation.
```
demolaf, line 3: Move this entry to `0.5.5-wip`. Merging main as-is auto-resolves it into the released `## 0.5.2` section, three releases back, so Remote Config would be missing from the next release notes.

.github/workflows/build.yml
```diff
@@ -287,6 +287,15 @@ jobs:
...
 
+      - name: Run Remote Config integration tests
+        env:
+          SERVICE_ACCOUNT: ${{ secrets.SERVICE_ACCOUNT }}
+        run: |
+          project_id="${SERVICE_ACCOUNT##*@}"
+          project_id="${project_id%%.iam.gserviceaccount.com}"
+          RC_TEST_PROJECT_ID="$project_id" \
+            dart test test/integration/remote_config/ --concurrency=1
+
```
demolaf, line 298: Gate this step on a single matrix entry, for example `if: matrix.dart-version == 'stable'`. `test-wif` runs stable and beta in parallel, so both jobs would publish and roll back the same live Remote Config project and race on ETags.

Run this step as `dart test test/integration/remote_config -P prod` and drop the `SERVICE_ACCOUNT` parsing. The `prod` preset arrives with main's `dart_test.yaml`, and `helpers.dart` already pins `projectId` to `dart-firebase-admin`, which is the project that tag documents, so `RC_TEST_PROJECT_ID` is not needed.

demolaf (changes requested):
Looks good, a couple of changes though.

PR #250, opened 2026-04-27
feat(app): make FirebaseApp.client public, added AppOptions.additionalScopes

### LINKED ISSUE #208: Make FirebaseApp.client fully public
**Is your feature request related to a problem? Please describe.**
Currently, I'm tinkering with `firebase_functions`.
These use `FirebaseApp` from `dart_firebase_admin` and expose the app to the consumer.

The `FirebaseApp` itself provides core firebase services which use the `package:googleapis_auth` `AuthClient`.

However, some of these services (e.g. `storage`) have a trimmed down API, which isn't sufficient for all use cases.

For instance, my current requirement is being able to upload a large file to Cloud Storage (several GBs) and fetch the file in ranges.
The API provided by default only supports direct data uploads and full data downloads, which isn't feasible for large files.

Whereas the `googleapis` package provides a more complete API, which supports streaming uploads.

Thus, it would be great if the `AuthClient` was exposed as part of the actual public API, so consumers can use it
without having to re-implement the same initialization logic `FirebaseApp` does under the hood, and so we don't have
to have separate clients for interacting with the same services.

**Describe the solution you'd like**
Remove the `@internal` annotation and make `AuthClient get client` part of the public API.

https://github.com/firebase/firebase-admin-dart/blob/4872123791ee8a49a17547cc7d4e95b40de886b5/packages/dart_firebase_admin/lib/src/app/firebase_app.dart#L138-L145

**Describe alternatives you've considered**
Silence the warning and use the internal API.

demolaf:
Closes #208.

`FirebaseApp.client` was `@internal`, forcing users who need an authenticated HTTP client (e.g. for advanced `googleapis` APIs not covered by this SDK) to either silence the warning and use the internal API, or re-implement the entire credential-initialization logic themselves.

### `FirebaseApp.client` — now public

Allows users to reuse the app's already-managed authenticated client for `googleapis` APIs, rather than spinning up a second client with a separate token lifecycle:

```dart
// Option A — reuse the app's managed client (same token refresh, same lifecycle)
final storageApi = StorageApi(await app.client);
final media = await storageApi.objects.get(
  bucket,
  objectName,
  downloadOptions: DownloadOptions.fullMedia,
);

// Option B — standalone client from a credential
final client = await Credential.fromApplicationDefaultCredentials()
    .createClient(['https://www.googleapis.com/auth/cloud-platform']);
```

`Credential.createClient()` and `FirebaseApp.client` are complementary: `createClient()` gives full scope control and separate lifecycle ownership; `client` shares the app's managed client.

### `AppOptions.additionalScopes` — new field

The SDK's default client always requests `cloud-platform` and `firebase` scopes. `additionalScopes` lets callers append extra scopes without having to manage their own client:

```dart
final app = FirebaseApp.initializeApp(
  options: AppOptions(
    credential: Credential.fromServiceAccount(File('sa.json')),
    additionalScopes: [
      'https://www.googleapis.com/auth/devstorage.full_control',
    ],
  ),
);
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
     timeout-minutes: 5
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@v5
 
       - name: Download addlicense
         run: |
@@ -118,7 +118,7 @@ jobs:
         with:
           fetch-depth: 2
 
-      - uses: actions/setup-java@v5
+      - uses: actions/setup-java@v4
         with:
           java-version: '21'
           distribution: 'temurin'
@@ -172,7 +172,7 @@ jobs:
 
       - name: Comment coverage on PR
         if: matrix.dart-version == 'stable' && github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false
-        uses: actions/github-script@v9
+        uses: actions/github-script@v7
         with:
           script: |
             const coverage = '${{ steps.coverage.outputs.coverage }}';
@@ -225,7 +225,7 @@ jobs:
       - name: Upload coverage to codecov
         if: matrix.dart-version == 'stable'
         continue-on-error: true
-        uses: codecov/codecov-action@v6
+        uses: codecov/codecov-action@v4
         with:
           files: coverage.lcov
           flags: unittests
@@ -259,7 +259,7 @@ jobs:
           sdk: ${{ matrix.dart-version }}
 
       - name: Authenticate to Google Cloud/Firebase
-        uses: google-github-actions/auth@v3
+        uses: google-github-actions/auth@v2
         with:
           workload_identity_provider: '${{ secrets.WORKLOAD_IDENTITY_PROVIDER }}'
           service_account: '${{ secrets.SERVICE_ACCOUNT }}'
--- a/packages/firebase_admin_sdk/CHANGELOG.md
+++ b/packages/firebase_admin_sdk/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 - Remove dependency on `package:equatable`.
 - Make `Query`, `CollectionReference`, `DocumentReference`, and `CollectionGroup` mockable.
+- `Credential.createClient(List<String> scopes)` — create an authenticated `AuthClient` directly 
+from a credential with custom scopes, without needing a `FirebaseApp` instance.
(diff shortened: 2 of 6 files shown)


Lyokone (approved):

packages/firebase_admin_sdk/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 - Remove dependency on `package:equatable`.
 - Make `Query`, `CollectionReference`, `DocumentReference`, and `CollectionGroup` mockable.
+- `Credential.createClient(List<String> scopes)` — create an authenticated `AuthClient` directly 
+from a credential with custom scopes, without needing a `FirebaseApp` instance.
+- AppOptions.additionalScopes` — append extra OAuth2 scopes to the SDK-managed HTTP client 
+without providing your own `AuthClient`.
```
kevmoo, line 8:
```suggestion
- `Credential.createClient(List<String> scopes)` — create an authenticated `AuthClient` directly 
  from a credential with custom scopes, without needing a `FirebaseApp` instance.
- AppOptions.additionalScopes` — append extra OAuth2 scopes to the SDK-managed HTTP client 
  without providing your own `AuthClient`.
```

.github/workflows/build.yml
```diff
@@ -259,7 +259,7 @@ jobs:
           sdk: ${{ matrix.dart-version }}
 
       - name: Authenticate to Google Cloud/Firebase
-        uses: google-github-actions/auth@v3
+        uses: google-github-actions/auth@v2
```
kevmoo, line 262: Revert these?

.github/workflows/build.yml
```diff
@@ -172,7 +172,7 @@ jobs:
 
       - name: Comment coverage on PR
         if: matrix.dart-version == 'stable' && github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false
-        uses: actions/github-script@v9
+        uses: actions/github-script@v7
```
kevmoo, line 175: Ditto...

.github/workflows/build.yml
demolaf: for some reason CI was not running, after dependabot updates to some actions https://github.com/firebase/firebase-admin-dart/actions/runs/24991582102, so I downgraded.

i'll investigate further which one is actually breaking CI

kevmoo: Annoying! Please look. I know firebase has some weird rules about non-Google action dependencies.

PR #251, opened 2026-04-27
PR #250 cleanup

demolaf:


demolaf:
@kevmoo the failing CI issue was caused by codecov-action@v6 not being allowed in the firebase org, reverted to codecov-action@v4.

kevmoo (approved):

PR #263, opened 2026-05-05
test: enable nan and inf tests

brianquinlan:


demolaf (approved):
Thanks! I saw this in the google_cloud_firestore_v1 migration PR too.

There's integration test for these too https://github.com/firebase/firebase-admin-dart/blob/1f3b0dd847ccdc8a0a9d110500a858221865c791/packages/google_cloud_firestore/test/integration/firestore_test.dart#L113

PR #269, opened 2026-05-07
fix: make UrlKeyFetcher http.Client injectable for testability

demolaf:
Makes the HTTP client used by `UrlKeyFetcher` injectable so that tests can intercept the Google public key certificate endpoint without making real network calls.

This is a prerequisite for the real token verification tests added in the `firebase_functions` SDK (firebase/firebase-functions-dart#91). Previously `UrlKeyFetcher` always used a freshly-created `http.Client`, making it impossible to intercept the certificate fetch in tests.


Lyokone (dismissed):
Could we have a separate PR for formatting changes so it's easier to review actual changes from this PR? (I guess it's related to a Flutter version bump?)

Ehesp (approved):

PR #281, opened 2026-06-01
feat: update firestore examples to include blob example

### LINKED ISSUE #280: [google_cloud_firestore] Blob support
**Is your feature request related to a problem? Please describe.**
I can read/write Blob using firestore in flutter. However I cannot read nor write Blob using google_cloud_firestore

**Describe the solution you'd like**
Add Blob support

**Describe alternatives you've considered**
Supporting Uint8List is an alternative

demolaf:


packages/firebase_admin_sdk/example/bin/example.dart
```diff
@@ -39,7 +39,7 @@ Future<void> main() async {
     // await projectConfigExample(admin);
 
     // Uncomment to run the firestore example
-    // await firestoreExample(admin);
+    await firestoreExample(admin);
```
Lyokone: Can we do a boolean at the top instead of this?

demolaf: how about this?

```dart
  try {
    if (_runFunctions) await functionsExample(admin);
    if (_runAuth) await authExample(admin);
    if (_runFirestore) await firestoreExample(admin);
    if (_runStorage) await storageExample(admin);
    if (_runMessaging) await messagingExample(admin);
    if (_runAppCheck) await appCheckExample(admin);
    if (_runSecurityRules) await securityRulesExample(admin);
  } finally {
    await admin.close();
  }
  ```

Lyokone (approved):

PR #290, opened 2026-06-24
feat: add toJson method to Timestamp

### LINKED ISSUE #289: Cannot return Timestamp from Cloud Functions as response
**Describe the bug**
Unlike node, attempting to send a `Timestamp` via `Response.ok` or `CallableResult` results in a error:

```
Converting object to an encodable object failed
```

**To Reproduce**

```dart
import 'package:firebase_functions/firebase_functions.dart';
import 'package:google_cloud_firestore/google_cloud_firestore.dart';

void main(List<String> args) {
  runFunctions((firebase) {
    firebase.https.onCall(name: 'getCurrentTimestamp', (_, _) async {
      return CallableResult({'timestamp': Timestamp.now()});
    });
  });
}

```

**Expected behavior**
A Map should be returned with seconds and nanoseconds of that Timestamp similar to node version.

OutdatedGuy:
## Related Issues

fixes #289 

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`).

- [x] I have updated the `CHANGELOG.md` of the relevant packages.
      Changelog files must be edited under the form:

  ```md
  ## Unreleased fix/major/minor

  - Description of your change. (thanks to @yourGithubId)
  ```

- [ ] If this contains new features or behavior changes,
      I have updated the documentation to match those changes.

### DIFF
--- a/packages/google_cloud_firestore/CHANGELOG.md
+++ b/packages/google_cloud_firestore/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - Updated `Transaction.delete` and `Transaction.update` type constraints to accept `DocumentReference<Object?>`. (thanks to @Levin-Me)
+- Made `Timestamp` encodable by adding `toJson` method. (thanks to @OutdatedGuy)
 
 ## 0.5.2
 
--- a/packages/google_cloud_firestore/lib/src/timestamp.dart
+++ b/packages/google_cloud_firestore/lib/src/timestamp.dart
@@ -197,4 +197,18 @@ final class Timestamp implements _Serializable {
   String toString() {
     return 'Timestamp(seconds=$seconds, nanoseconds=$nanoseconds)';
   }
+
+  /// Converts this Timestamp to a JSON representation.
+  ///
+  /// Returns a Map\<String, int\> containing the seconds and nanoseconds.
+  ///
+  /// Example:
+  /// ```dart
+  /// final timestamp = Timestamp(seconds: 1234567890, nanoseconds: 123456789);
+  /// final json = timestamp.toJson();
+  /// print(json); // {'_seconds': 1234567890, '_nanoseconds': 123456789}
+  /// ```
+  Map<String, int> toJson() {
+    return {'_seconds': seconds, '_nanoseconds': nanoseconds};
+  }
 }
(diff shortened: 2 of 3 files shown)


demolaf (changes requested):
I think this should be in the firebase_functions package itself, specifically CallableResponse in `callable.dart` https://github.com/firebase/firebase-functions-dart/blob/main/lib/src/https/callable.dart#L419-L457.

Also, `Timestamp.toJson()` should not be used internally since the bundle serialization format requires a different set of map keys (seconds/nanos).

OutdatedGuy:
hi @demolaf, the Timestamp in node.js is natively encodable without depending on `firebase-functions`

```ts
import { Timestamp } from "firebase-admin/firestore";

const now = Timestamp.now();

console.log(JSON.stringify(now)); // {"_seconds":1782405004,"_nanoseconds":354000000}

```

Having a native `toJson` method would be better for custom dart backends not using firebase cloud functions.

---

Also, the `encode` method you linked seems to used only for streaming responses, not for `CallableResult`.

`CallableResult` uses `dart:convert`'s `jsonEncode` method:

https://github.com/firebase/firebase-functions-dart/blob/bf539c4863b2ba90c8d6e0ef166119ce083dca0e/lib/src/https/callable.dart#L45-L49

demolaf:
@Lyokone @lahirumaramba can I get your thoughts on this? 

my main concern here is `Timestamp.toJson()` if added can be called via `jsonEncode` which is a risk e.g. `BundleMetadata` https://github.com/firebase/firebase-admin-dart/blob/main/packages/google_cloud_firestore/lib/src/bundle.dart#L61-L72 uses `Timestamp` but in the future can be mistaken and `Timestamp.toJson()` used.

firebase functions node never had this concern because it doesn't declare a `Timestamp.toJson()`.

@OutdatedGuy encode is not only for streaming responses, see node reference https://github.com/firebase/firebase-functions/blob/master/src/common/providers/https.ts#L919-L927 it's job here is to convert types into json compatible values/throws for incompatible json values and that's actually a bug that it is not being called in `CallableResult.toResponse()`.

the idea here is to let firebase functions dart handle functions specific work instead of firebase admin since it can't defend against node's default behavior.

Lyokone (approved):

demolaf (approved):

PR #292, opened 2026-06-29
feat: add support for Firestore Pipeline

Lyokone:
Add support for Firestore Pipeline

How to use
```dart
final snapshot = await firestore
    .pipeline()
    .collection('books')
    .where(Expression.field('active').equalValue(true))
    .sort([Expression.field('price').ascending()])
    .select([
      Expression.field('title'),
      Expression.field('price'),
      Expression.field('title').toUpperCase().as('upperTitle'),
      Expression.field('tags').arrayLength().as('tagCount'),
    ])
    .limit(10)
    .execute();
for (final result in snapshot.results) {
  print(result.data());
}
```

#### Aggregates

Aggregate stages use aliased aggregate expressions:

```dart
final snapshot = await firestore
    .pipeline()
    .collection('books')
    .where(Expression.field('active').equalValue(true))
    .aggregate([
      Expression.field('price').sum().as('totalPrice'),
      Expression.field('rating').average().as('averageRating'),
      PipelineFunctions.count().as('bookCount'),
    ])
    .execute();
final data = snapshot.results.single.data();
print(data);
```

#### Expressions

Use `Expression.field`, `Expression.constant`, and `Expression.variable` to
build expressions. Most helpers are also available as fluent methods:

```dart
final expression = Expression.field('createdAt')
    .timestampSubtract('day', 7)
    .timestampToUnixSeconds()
    .as('createdSeconds');
```


packages/google_cloud_firestore/lib/src/pipeline.dart
```diff
@@ -0,0 +1,2173 @@
...
+}
+
+/// Creates a descending Pipeline ordering.
+PipelineOrdering descending(Object expression) {
+  return PipelineOrdering._('descending', expression);
+}
+
+/// The starting point for constructing Firestore Pipeline operations.
+@immutable
+final class PipelineSource {
```
demolaf, line 1125: we are missing `createFrom(query)` https://github.com/googleapis/google-cloud-node/blob/69d6d21cebc6237d15be9d2adacf0f3a63fd8f1f/handwritten/firestore/dev/src/pipelines/pipelines.ts#L265

packages/google_cloud_firestore/lib/src/pipeline.dart
```diff
@@ -0,0 +1,2173 @@
...
+  }
+
+  /// CHAR_LENGTH string function.
+  static PipelineExpression charLength(Object? value) {
+    return _expr('char_length', [value]);
+  }
+
+  /// STARTS_WITH string function.
+  static PipelineBooleanExpression startsWith(Object? value, Object? prefix) {
+    return _bool('starts_with', [value, prefix]);
```
demolaf: Since `value` is typed as `Object?`, we should check at runtime whether it was passed as a `String` and wrap it with `field(value)` so it's encoded as a field reference.

Without that check, the string is encoded as a literal value instead of a field reference — so the expression compares against the fixed text `'title'`, not the value of the `title` field.

Example (Bug) — this looks like it filters on the `title` field, but instead compares the literal text `"title"` against `"Harry"`:

```dart
await firestore
    .pipeline()
    .collection('books')
    .where(PipelineFunctions.startsWith('title', 'Harry'))
    .execute();
```

This problem also exists in other methods across `PipelineFunctions` e.g. `equal`, `lessThan`, `arrayContains`, `mapGet`, and others

packages/google_cloud_firestore/lib/src/pipeline.dart
```diff
@@ -0,0 +1,2173 @@
...
+    return _expr('byte_length', [value]);
+  }
+
+  /// CHAR_LENGTH string function.
+  static PipelineExpression charLength(Object? value) {
+    return _expr('char_length', [value]);
+  }
+
+  /// STARTS_WITH string function.
+  static PipelineBooleanExpression startsWith(Object? value, Object? prefix) {
```
demolaf: we should rename `value` to `fieldName` which is more descriptive and matches node.

packages/google_cloud_firestore/lib/src/pipeline.dart
```diff
@@ -0,0 +1,2173 @@
...
+
+  /// JOIN function.
+  static PipelineExpression join(Object? array, [Object? separator]) {
+    return _expr('join', [array, ..._optionalArg(separator)]);
+  }
+
+  /// EQUAL comparison function.
+  static PipelineBooleanExpression equal(Object? left, Object? right) {
+    return _bool('equal', [left, right]);
+  }
```
demolaf, line 548: same issue here, we should check if `left is String` and wrap with `field(left)`.

- In node, `fieldOrExpression` handles converting `left` to field if a string was passed:
https://github.com/googleapis/google-cloud-node/blob/main/handwritten/firestore/dev/src/pipelines/expression.ts#L5061-L5068

- pipeline node tests (validates bug): 
https://github.com/googleapis/google-cloud-node/blob/main/handwritten/firestore/dev/system-test/pipeline.ts#L1128-L1144

demolaf (changes requested):
`PipelineFunctions` accepts `Object?` but only `Expression.field()` actually works without a fieldOrExpression-style check — untested since every call site already used `field()`. We should either commit to requiring `Expression.field() `explicitly or handle the coercion like Node's [fieldOrExpression](https://github.com/googleapis/google-cloud-node/blob/main/handwritten/firestore/dev/src/pipelines/pipeline-util.ts#L697-L705).

packages/google_cloud_firestore/lib/src/pipeline.dart
dackers86: Now added by Guillaume.

dackers86: @demolaf This feature has now been added, could you please re-review.

demolaf (approved):

Issue #297, opened 2026-07-06
Concurrent request execution is exponentially slow


OutdatedGuy:
**Describe the bug**
Concurrently executing thousands of queries is super slow in dart. For below code it took around ~2s for nodejs version to finish executing and around ~57s for dart version.

**To Reproduce**

The code should call production (actual firestore) APIs and not a local emulator to see the issue.

Dart (Super Slow):
```dart
import 'package:firebase_admin_sdk/firebase_admin_sdk.dart';
import 'package:firebase_functions/firebase_functions.dart';

void main() {
  runFunctions((firebase) {
    firebase.https.onRequest(name: 'helloWorld', (request) async {
      final arr = List.generate(3000, (index) => index);

      final firestore = FirebaseApp.instance.firestore();

      await Future.wait(
        arr.map((index) {
          return firestore
              .collection('numbers')
              .where('number', .equal, index)
              .limit(1)
              .get();
        }),
        eagerError: true,
      );

      return .ok('Yay! Client Exception did not occur.');
    });
  });
}

```

Typescript (Fast):
```ts
import {initializeApp} from "firebase-admin/app";
import {getFirestore} from "firebase-admin/firestore";
import {onRequest} from "firebase-functions/https";
initializeApp();

export const helloWorld = onRequest(async (_request, response) => {
  const arr = Array.from({length: 3000}, (_, index) => index);

  const firestore = getFirestore();

  await Promise.all(
    arr.map((index) => {
      return firestore
        .collection("numbers")
        .where("number", "==", index)
        .limit(1)
        .get();
    }),
  );

  response.status(200).send("Yay! Client Exception did not occur.");
});

```

**Expected behavior**
The speed of executing thousands of queries concurrently in dart should be as fast as if not faster than nodejs version.

brianquinlan:
@lahirumaramba wrote:

> I also believe the delay is caused by resource exhaustion with HTTP/1. the TypeScript version is likely faster because @google-cloud/firestore (Node.js) uses gRPC over HTTP/2 with multiplexing.
> 
> The optimal solution would be to implement HTTP/2 multiplexing in Dart. Do we currently use HTTP/2 for Messaging? Maybe we can reuse some of that logic here.
> 
[...]

PR #298, opened 2026-07-06
feat(firestore): add maxConnectionsPerHost to Settings

mosuem:
[WARNING: AI generated, as inspiration only, not meant to land]

Allows configuring the underlying HTTP client's maxConnectionsPerHost to improve performance of concurrent requests when using the REST API.

## Related Issues

fixes #your-issue-number

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`).

- [ ] I have updated the `CHANGELOG.md` of the relevant packages.
      Changelog files must be edited under the form:

  ```md
  ## Unreleased fix/major/minor

  - Description of your change. (thanks to @yourGithubId)
  ```

- [ ] If this contains new features or behavior changes,
      I have updated the documentation to match those changes.

### DIFF
--- a/packages/google_cloud_firestore/lib/src/firestore_http_client.dart
+++ b/packages/google_cloud_firestore/lib/src/firestore_http_client.dart
@@ -13,12 +13,14 @@
 // limitations under the License.
 
 import 'dart:async';
+import 'dart:io';
 
 import 'package:google_cloud/constants.dart' as google_cloud;
 import 'package:google_cloud/google_cloud.dart' as google_cloud;
 import 'package:google_cloud_firestore_v1/firestore.dart' as firestore_v1;
 import 'package:googleapis_auth/auth_io.dart' as googleapis_auth;
 import 'package:http/http.dart';
+import 'package:http/io_client.dart';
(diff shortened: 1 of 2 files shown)


packages/google_cloud_firestore/lib/src/firestore_http_client.dart
```diff
@@ -156,22 +158,30 @@ class FirestoreHttpClient {
 
   /// Creates the appropriate HTTP client based on emulator configuration.
   Future<googleapis_auth.AuthClient> _createClient() async {
+    Client? baseClient;
```
brianquinlan, line 161: Do we allow users to inject their own `Client`?

Do we want to expose a `dart:io`-specific bit of configuration here?

demolaf:
thanks @mosuem `maxConnectionsPerHost` would just limit the number of concurrent requests that can be executed and queue the rest, which is significantly slower but http/2 with pooled connections matches node's implementation.

See my comment - https://github.com/firebase/firebase-admin-dart/issues/297#issuecomment-5057056950

Closing this PR.

PR #302, opened 2026-07-10
feat(functions): Implement Task Queue Scopes

inlined:
This pull request implements Task Queue Scopes in the Functions service. Task Queue Scopes allow developers to specify the scope of a function in a task queue (such as targeting `current`, `global`, `extension`, or `kit` scopes) using a new `FunctionScope` parameter in `taskQueue()`.

This implementation is based on the design patterns and logic from the Node.js SDK (see [PR #3210](https://github.com/firebase/firebase-admin-node/pull/3210)), with the following modifications:
* Cases where the SDK targets its own extension have been removed because Firebase Extensions do not use the Dart runtime.
* The error handling flows have been flattened for better readability.

Apologies if I missed any simple Dart conventions or errors; I am still getting familiar with Dart.

#### Testing Strategy
Added unit tests in `test/unit/functions/functions_test.dart` to verify the different `FunctionScope` behaviors, warnings, and fallback/retry behavior. All 74 unit tests and `dart analyze` pass successfully.

Context Sources Used:
- Node.js SDK Implementation Reference: [firebase-admin-node PR #3210](https://github.com/firebase/firebase-admin-node/pull/3210)

### DIFF
--- a/packages/firebase_admin_sdk/lib/src/app/environment.dart
+++ b/packages/firebase_admin_sdk/lib/src/app/environment.dart
@@ -174,4 +174,11 @@ abstract class Environment {
         Zone.current[envSymbol] as Map<String, String>? ?? Platform.environment;
     return env[cloudTasksEmulatorHost];
   }
+
+  /// Gets the Kit Instance ID if running within a kit.
+  static String? getKitInstanceId() {
+    final env =
+        Zone.current[envSymbol] as Map<String, String>? ?? Platform.environment;
+    return env['FIREBASE_KIT_INSTANCE_ID'];
+  }
 }
--- a/packages/firebase_admin_sdk/lib/src/functions/functions_api.dart
+++ b/packages/firebase_admin_sdk/lib/src/functions/functions_api.dart
@@ -174,3 +174,59 @@ class TaskOptions {
   /// Contains experimental features that may change in future releases.
   final TaskOptionsExperimental? experimental;
 }
+
+/// Represents the scope of a function in a task queue.
+sealed class FunctionScope {
+  const FunctionScope._();
+
+  /// Targets a function co-deployed in the same codebase/deployment context.
+  const factory FunctionScope.current() = _CurrentFunctionScope;
+
+  /// Targets a global function (e.g. without kit or extension namespacing).
+  const factory FunctionScope.global() = _GlobalFunctionScope;
+
+  /// Targets a function deployed within a specific Firebase Extension.
+  const factory FunctionScope.extension(String extensionId) =
+      _ExtensionFunctionScope;
+}
+
+class _GlobalFunctionScope extends FunctionScope {
+  const _GlobalFunctionScope() : super._();
+
+  @override
+  String toString() => 'FunctionScope.global()';
+}
+
+class _ExtensionFunctionScope extends FunctionScope {
+  const _ExtensionFunctionScope(this.extensionId) : super._();
+
+  final String extensionId;
+
+  @override
+  String toString() => 'FunctionScope.extension($extensionId)';
+}
+
+class _KitFunctionScope extends FunctionScope {
+  const _KitFunctionScope(this.kit) : super._();
+
+  final String kit;
+
+  @override
+  String toString() => 'FunctionScope.kit($kit)';
+}
+
+class _CurrentFunctionScope extends FunctionScope {
+  const _CurrentFunctionScope() : super._();
+
+  @override
+  String toString() => 'FunctionScope.current()';
+}
+
+class _ExtensionOrKitFunctionScope extends FunctionScope {
+  const _ExtensionOrKitFunctionScope(this.instance) : super._();
+
+  final String instance;
+
+  @override
+  String toString() => 'FunctionScope.extensionOrKit($instance)';
+}
--- a/packages/firebase_admin_sdk/lib/src/functions/task_queue.dart
+++ b/packages/firebase_admin_sdk/lib/src/functions/task_queue.dart
@@ -51,7 +51,7 @@ class TaskQueue {
         print(
           'Targeting your own extension or kit no longer requires a second parameter, '
           'which can have performance implications. Please change the call '
-          "taskQueue('$_functionName', '$instance') to taskQueue('$_functionName') "
+          "taskQueue('$_functionName', extensionId: '$instance') to taskQueue('$_functionName') "
           "or taskQueue('$_functionName', scope: FunctionScope.current())",
         );
       }
@@ -153,7 +153,7 @@ class TaskQueue {
     print(
       'Targeting kit $instance with the legacy extensions API, '
       'which has performance implications. Please change the call '
-      "taskQueue('$functionName', '$instance') to taskQueue('$functionName')",
+      "taskQueue('$functionName', extensionId: '$instance') to taskQueue('$functionName')",
     );
   }
 }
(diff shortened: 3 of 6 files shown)


inlined:
Note: Node is forwards compatible with kits, but doesn't actually expose the type for kit. Any guidance on how to hide kits' existence until a later release is appreciated.

demolaf:
> Note: Node is forwards compatible with kits, but doesn't actually expose the type for kit. Any guidance on how to hide kits' existence until a later release is appreciated.

@inlined `_KitFunctionScope` is already private so dropping `FunctionScope.kit` removes it from the public API and where `FunctionScope.kit` is used can instead be replaced with `_KitFunctionScope` directly.

**OR**

annotate `FunctionScope.kit` with `internal` from `package:meta` (compile time warning) which keeps the implementation working as-is.

packages/firebase_admin_sdk/lib/src/functions/task_queue.dart
```diff
@@ -23,18 +23,44 @@ class TaskQueue {
...
+      if (kitInstanceId != null &&
+          kitInstanceId.isNotEmpty &&
+          kitInstanceId == instance) {
+        _scope = FunctionScope.kit(instance);
+        print(
+          'Targeting your own extension or kit no longer requires a second parameter, '
+          'which can have performance implications. Please change the call '
+          "taskQueue('$_functionName', '$instance') to taskQueue('$_functionName') "
+          "or taskQueue('$_functionName', scope: FunctionScope.current())",
+        );
```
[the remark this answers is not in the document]
inlined: The SDK does not currently use a logging package in its dependencies, and uses standard `print()` statements for warnings (e.g. in `environment.dart`). We continued using `print()` here to remain consistent. Note that the own-extension warnings/checks have also been removed per feedback.

packages/firebase_admin_sdk/lib/src/functions/task_queue.dart
```diff
@@ -74,7 +120,45 @@ class TaskQueue {
...
+        // Only upgrade the stateful scope to kit if the retry request succeeds
+        _scope = tempKitScope;
+        _logFallbackWarning(_functionName, currentScope.instance);
+      } on FirebaseFunctionsAdminException catch (retryErr) {
+        if (retryErr.errorCode != FunctionsClientErrorCode.notFound) {
+          rethrow;
+        }
+      }
+    }
+  }
```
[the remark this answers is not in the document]
inlined: Implemented as suggested to simplify the delete method and flatten the nested try-catch blocks.

packages/firebase_admin_sdk/lib/src/functions/task_queue.dart
```diff
@@ -74,7 +120,45 @@ class TaskQueue {
...
+
+  void _logFallbackWarning(String functionName, String instance) {
+    print(
+      'Targeting kit $instance with the legacy extensions API, '
+      'which has performance implications. Please change the call '
+      "taskQueue('$functionName', '$instance') to "
+      "taskQueue('$functionName', scope: FunctionScope.kit('$instance'))",
+    );
```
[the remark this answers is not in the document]
inlined: Since the package does not use a logging library in its dependencies and relies on standard `print()` statements for existing warnings, we kept using `print()` here to remain consistent.

packages/firebase_admin_sdk/lib/src/app/environment.dart
```diff
@@ -174,4 +174,11 @@ abstract class Environment {
         Zone.current[envSymbol] as Map<String, String>? ?? Platform.environment;
     return env[cloudTasksEmulatorHost];
   }
+
+  /// Gets the Kit Instance ID if running within a kit.
+  static String? getKitInstanceId() {
+    final env =
+        Zone.current[envSymbol] as Map<String, String>? ?? Platform.environment;
+    return env['FIREBASE_KIT_INSTANCE_ID'];
+  }
```
demolaf, line 183: maybe we can a firebase doc link when this is released, so we know how/from where this `'FIREBASE_KIT_INSTANCE_ID'` is injected?

demolaf (approved):
Looks good 👍

packages/firebase_admin_sdk/lib/src/app/environment.dart
inlined: Definitely worth considering. It is a reserved env var; I could see us documenting it, though with this API it shouldn't actually be necessary to set or read directly I believe.

packages/firebase_admin_sdk/lib/src/functions/functions_api.dart
```diff
@@ -174,3 +174,59 @@ class TaskOptions {
...
+  /// Targets a function co-deployed in the same codebase/deployment context.
+  const factory FunctionScope.current() = _CurrentFunctionScope;
+
+  /// Targets a global function (e.g. without kit or extension namespacing).
+  const factory FunctionScope.global() = _GlobalFunctionScope;
+
+  /// Targets a function deployed within a specific Firebase Extension.
+  const factory FunctionScope.extension(String extensionId) =
+      _ExtensionFunctionScope;
+}
```
lahirumaramba, line 191: Just clarifying, we didn't include kit here because we don't plan to expose it in the public api, yet?
```dart
  /// Targets a function deployed within a specific Firebase Kit.
  const factory FunctionScope.kit(String kit) = _KitFunctionScope;
```

packages/firebase_admin_sdk/lib/src/functions/task_queue.dart
```diff
@@ -23,18 +23,44 @@ class TaskQueue {
...
+      final instance = (_scope as _ExtensionOrKitFunctionScope).instance;
+      final kitInstanceId = Environment.getKitInstanceId();
+      if (kitInstanceId != null &&
+          kitInstanceId.isNotEmpty &&
+          kitInstanceId == instance) {
+        _scope = _KitFunctionScope(instance);
+        print(
+          'Targeting your own extension or kit no longer requires a second parameter, '
+          'which can have performance implications. Please change the call '
+          "taskQueue('$_functionName', '$instance') to taskQueue('$_functionName') "
```
lahirumaramba: nit: should this be a named argument? `extensionId: '$instance'`

packages/firebase_admin_sdk/lib/src/functions/task_queue.dart
```diff
@@ -74,7 +120,40 @@ class TaskQueue {
...
+      }
+      // Swallow notFound errors as delete is idempotent.
+    }
+  }
+
+  void _logFallbackWarning(String functionName, String instance) {
+    print(
+      'Targeting kit $instance with the legacy extensions API, '
+      'which has performance implications. Please change the call '
+      "taskQueue('$functionName', '$instance') to taskQueue('$functionName')",
```
lahirumaramba: nit: should this be a named argument? `extensionId: '$instance'`

lahirumaramba (approved):

packages/firebase_admin_sdk/lib/src/functions/functions_api.dart
inlined: Yes, Kit is not a public feature (or name) yet, so we're keeping it out of the public contract

packages/firebase_admin_sdk/lib/src/functions/task_queue.dart
inlined: Yes, you're right. extensionId is a named parameter in Dart. I've updated the warning print messages to suggest the named parameter format.

packages/firebase_admin_sdk/lib/src/functions/task_queue.dart
inlined: Updated to suggest using the named parameter format.

PR #305, opened 2026-07-23
fix(firestore): use pooled HTTP/2 connections by default to fix slow/failing concurrent requests

### LINKED ISSUE #297: Concurrent request execution is exponentially slow

demolaf:
Fixes #297.

Dart's Firestore client used `dart:io`'s `HttpClient`, which only speaks HTTP/1.1, so every concurrent request opened its own TCP+TLS connection. At high concurrency this caused connection-storm failures (`ClientException`, `HandshakeException`) and made Dart much slower than Node, whose Firestore client multiplexes concurrent operations over a small pool of HTTP/2 connections.

Firestore traffic now goes over `package:http2`'s `Http2Client`, which packs requests onto shared connections under a 100-stream cap and dials another once those fill. It speaks only HTTP/2, so a small wrapper keeps the endpoints that are still plain HTTP on HTTP/1.1: the emulator, `Settings.ssl: false`, and the credential sources `googleapis_auth` reaches through the same base client, such as the GCE metadata server.

## Benchmark

| Concurrency | Run 1 | Run 2 | Run 3 |
|---|---|---|---|
| 3000 | 2438.35ms | 1602.67ms | 1245.91ms |
| 5000 | 2460.53ms | 2199.92ms | 2111.67ms |
| 10000 | 7195.02ms | 1518.93ms | 2180.65ms |

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -234,7 +234,7 @@ jobs:
       - name: Upload coverage to codecov
         if: matrix.dart-version == 'stable'
         continue-on-error: true
-        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f
+        uses: codecov/codecov-action@v4
         with:
           files: coverage.lcov
           flags: unittests
--- a/packages/google_cloud_firestore/CHANGELOG.md
+++ b/packages/google_cloud_firestore/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Updated `Transaction.delete` and `Transaction.update` type constraints to accept `DocumentReference<Object?>`. (thanks to @Levin-Me)
 - Made `Timestamp` encodable by adding `toJson` method. (thanks to @OutdatedGuy)
 - Update dependency `googleapis_auth: ^2.3.3` to fix `auth/insufficient-permission` errors with Application Default Credentials that have no quota project set.
+- Fixed slow and failing requests under high concurrency by pooling shared HTTP/2 connections by default instead of opening a new HTTP/1.1 connection per request.
 
 ## 0.5.2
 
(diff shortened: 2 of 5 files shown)


brianquinlan:
@demolaf

Would it make sense for the `Client` interface to live in `package:http2`?

demolaf:
@brianquinlan this makes sense, do we land this first and open a new issue/PR on `dart-lang/http` repo or vice versa?

brianquinlan:
I would love to see a PR on the dart-lang/http repo. I'm going to be out next week so maybe @natebosch can review

Lyokone (approved):

PR #306, opened 2026-07-23
fix(ci): use allow-listed codecov-action@v4 ref instead of a SHA pin

demolaf:


demolaf:
The org's Actions allow-list only permits the literal ref codecov/codecov-action@v4, but this repo's zizmor scan mandates every action be SHA-pinned — and confirmed empirically, the allow-list rejects any SHA for this action, even the exact commit v4 points to.

These two policies directly conflict for this one action; someone with access to either the org allow-list or the zizmor mandatory-checks config needs to grant an exception.

Lyokone (approved):

PR #308, opened 2026-07-23
fix(admin): attach usage-tracking headers to Firestore requests

demolaf:
`Firestore` requests never carried the SDK's usage-tracking headers (`X-Firebase-Client`, `X-Goog-Api-Client`) — `FirestoreHttpClient` builds its own HTTP client independently of `FirebaseApp.client`/`FirebaseUserAgentClient`, so it was never wrapped.

Uses the new `Settings.headers` from #307 to attach the same headers, and extracts a shared `firebaseUserAgentHeaders` helper so the values aren't duplicated between `FirebaseUserAgentClient` and `Firestore`. Depends on #307 merging first.

### DIFF
--- a/packages/firebase_admin_sdk/CHANGELOG.md
+++ b/packages/firebase_admin_sdk/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 0.5.4-wip
 
+- Fixed `Firestore` requests not carrying the SDK's usage-tracking headers (`X-Firebase-Client`, `X-Goog-Api-Client`).
 - Update dependency `googleapis_auth: ^2.3.3` to fix `auth/insufficient-permission`
   errors with Application Default Credentials that have no quota project set.
 
--- a/packages/firebase_admin_sdk/CHANGELOG.md
+++ b/packages/firebase_admin_sdk/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.5.4-wip
 
-- Fixed `Firestore` requests not carrying the SDK's usage-tracking headers (`X-Firebase-Client`, `X-Goog-Api-Client`).
+- Fixed `Firestore` requests not carrying the SDK's usage-tracking headers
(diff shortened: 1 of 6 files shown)


packages/firebase_admin_sdk/lib/src/app/firebase_user_agent_client.dart
```diff
@@ -32,9 +32,7 @@ class FirebaseUserAgentClient extends BaseClient
 
   @override
   Future<StreamedResponse> send(BaseRequest request) {
-    request.headers['X-Firebase-Client'] = 'fire-admin-dart/$packageVersion';
-    request.headers['X-Goog-Api-Client'] =
-        'gl-dart/$dartVersion fire-admin/$packageVersion';
+    request.headers.addAll(firebaseUserAgentHeaders);
```
brianquinlan, line 35: Is this semantically different than what was there before?

packages/firebase_admin_sdk/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## 0.5.4-wip
 
+- Fixed `Firestore` requests not carrying the SDK's usage-tracking headers (`X-Firebase-Client`, `X-Goog-Api-Client`).
```
brianquinlan: Is this meant to be wrapped at 80 columns.

brianquinlan (approved):

packages/firebase_admin_sdk/CHANGELOG.md
demolaf: Good catch, fixed — wrapped to match the entry below it.

packages/firebase_admin_sdk/lib/src/app/firebase_user_agent_client.dart
demolaf: No, same values — just centralized into one shared constant instead of duplicated at each call site.

PR #318, opened 2026-08-31
fix(firestore): align the Pipeline API with the Node Admin SDK

### LINKED ISSUE #317: Firestore Pipeline: follow-ups from the #292 review
Follow-up work from the review of #292 (merged as 9652fc9). Finding A from that review — the `documents()` source stage encoding — was fixed in-branch before merge and is confirmed green against the live database. The items below are everything that was deliberately deferred.

Line references are against `main` at 9652fc9.

**Status: all items addressed in #318.** Item 6 was wrong and has been withdrawn; everything else is implemented, tested and green. Item 9 was added while doing the work.

| # | Item | Status |
| --- | --- | --- |
| 1 | Transaction support incomplete and unreachable | Done |
| 2 | Source stages skip database validation | Done — also compares project |
| 3 | `explainStats` exposes a generated proto type | Done |
| 4 | 14 `*Value`/`*Number` synonyms | Done |
| 5 | `data()` nullable but never null | Done |
| 6 | Missing `between`/`matches`/`arrayLastIndexOf` | Withdrawn — the finding was wrong |
| 7 | CHANGELOG describes intra-PR history | Done |
| 8 | `withOptions()` takes an untyped `Map` | Done |
| 9 | Expression names diverge from the Node SDK | Done |

---

## 1. Pipeline transaction support is incomplete and unreachable — DONE

**Type:** bug

Two halves of the same gap:

1. `Pipeline.execute()` reads `chunk.transaction` and surfaces it as `PipelineSnapshot.transaction` (`pipeline.dart:1469`), but the request never sets `ExecutePipelineRequest.newTransaction` — the only thing that makes the backend return one. The field is permanently `null`.
2. `execute({String? transaction})` (`pipeline.dart:1438`) takes a raw base64 transaction ID, but `Transaction` exposes no public accessor for its ID, and there is no `Transaction.executePipeline(...)` alongside the existing `getQuery` / `getAggregateQuery` / `getAll`. There is no supported way for a caller to obtain a valid value for this parameter.

Net effect: pipelines cannot be run inside a transaction at all, and `PipelineSnapshot.transaction` is dead surface.

**Fix:** add `Transaction.executePipeline(Pipeline)` following the `getQuery` pattern, and either wire up a `newTransaction` request option or drop `PipelineSnapshot.transaction`.

---

## 2. Source stages don't validate the database the way `createFrom` does — DONE

**Type:** bug

`PipelineSource.createFrom()` calls `_validateSameDatabase` (`pipeline.dart:1195`) and throws a clear `ArgumentError` on a cross-database query. These three accept references and pipelines from a different `Firestore` instance with no check, silently producing a request that targets the wrong database:

- `PipelineSource.collectionReference()` — `pipeline.dart:1150`
- `PipelineSource.documents()` — `pipeline.dart:1157`
- `Pipeline.union()` — `pipeline.dart:1351`

**Fix:** apply `_validateSameDatabase` consistently across all three.

---

## 3. `PipelineSnapshot.explainStats` exposes an unexported generated proto type — DONE

**Type:** enhancement — **public API, best fixed before release**

`PipelineSnapshot.explainStats` is typed `firestore_v1.ExplainStats?` (`pipeline.dart:1892`). `firestore_v1` is not exported from `google_cloud_firestore.dart`, so consumers cannot name the type or write a variable declaration for it.

Everywhere else the package hand-wraps this: `ExplainMetrics`, `ExecutionStats`, and `PlanSummary` in `lib/src/query_profile.dart`, all exported, all used by `Query`, `AggregateQuery`, and `VectorQuery`.

**Fix:** add a wrapper in `query_profile.dart` mirroring `ExplainMetrics._fromProto`, and export it.

---

## 4. 14 `*Value` / `*Number` methods are exact synonyms of their base methods — DONE

**Type:** enhancement — **public API, best fixed before release**

```dart
PipelineBooleanExpression equalValue(Object? value) => equal(value);   // pipeline.dart:1925
PipelineExpression addNumber(num other) => add(other);
```

All 14 delegate verbatim: `equalValue`, `notEqualValue`, `lessThanValue`, `lessThanOrEqualValue`, `greaterThanValue`, `greaterThanOrEqualValue`, `addNumber`, `subtractNumber`, `multiplyNumber`, `divideNumber`, `moduloNumber`, `arrayContainsValue`, `ifAbsentValue`, `ifErrorValue`.

These names come from the Firebase JS Web SDK, where `eqValue` forces `constant(value)` and `eq` accepts an `Expr`. That distinction is the entire reason the pair exists. Here both paths are identical, so passing an expression to `equalValue` gets expression semantics — the opposite of what the name promises. The Node Admin SDK has no such methods at all.

Same duplication in:
- `PipelineExpression.alias()` / `as()` — `pipeline.dart:1914`
- `PipelineSnapshot.results` / `result` — `pipeline.dart:1883`

**Fix:** either wrap in `Expression.constant()` so the names mean something, or drop all 14 plus the two aliases.

---

## 5. `PipelineResult.data()` is typed nullable but never returns null — DONE

**Type:** enhancement

`pipeline.dart:1831`:

```dart
DocumentData? data() => Map.unmodifiable(_data);
```

The return is unconditional, forcing callers into a null check that can never fire. It also allocates a fresh unmodifiable view on every call, which matters when walking a large result set.

**Fix:** make the return non-nullable and build the unmodifiable map once in the constructor.

---

## 6. ~~Missing expression functions: `between`, `matches`, `arrayLastIndexOf`~~ — WITHDRAWN, this was wrong

**Type:** ~~enhancement, good first issue~~ — no action needed

This finding was incorrect on all three counts:

- `between` and `matches` are **commented out** in Node's `expression.ts`. They are not part of its public API. The original comparison grepped `new FunctionExpression('...')` across the file without stripping comments, so it matched dead code.
- `arrayLastIndexOf` **already exists** in this package, in both static and fluent form, with exactly the implementation that was proposed. The comparison ran at the level of backend function names, and `arrayIndexOf`/`arrayLastIndexOf` both emit `array_index_of`, so the method-level gap was inferred rather than checked.

Re-running the comparison against Node's live code with comments stripped: every backend function Node exposes is already covered here. Nothing to do.

---

## 8. Pipeline options were an untyped `Map`, so callers had to know proto field names — DONE

**Type:** enhancement — public API

Requesting explain stats meant writing server-side field names by hand, where a wrong key failed silently:

```dart
.withOptions({
  'explain_options': {'mode': 'analyze', 'output_format': 'text'},
})
```

**What shipped is not the shape originally proposed here.** The first proposal kept `Pipeline.withOptions()` as a builder and added `withRawOptions()` beside it. A later check against Node showed `withOptions` does not exist there at all — options are passed to `execute()`, with `rawOptions` as a field on the same object rather than a separate method. Node's public `PipelineExecuteOptions` is:

```ts
{ indexMode?: 'recommended';
  explainOptions?: { mode?: 'execute' | 'analyze'; outputFormat?: 'text' };
  rawOptions?: {...} }
```

So the builder methods were dropped and the options moved onto `execute()`:

```dart
await firestore.pipeline().collection('books').execute(
  indexMode: PipelineIndexMode.recommended,
  explain: const PipelineExplainOptions(
    mode: PipelineExplainMode.analyze,
    outputFormat: PipelineExplainOutputFormat.text,
  ),
  rawOptions: {'some_new_option': true},  // takes precedence
);
```

with `PipelineIndexMode { recommended }`, `PipelineExplainMode { execute, analyze }` and `PipelineExplainOutputFormat { text }`. Note `PipelineExplainMode` has **two** values, not the three first proposed here — `'explain'` appears in Node's internal `structured-pipeline.ts` type but not in its public `types/firestore.d.ts`, so it is not exposed.

A side effect worth recording: because options are no longer pipeline state, `Pipeline` lost its `_options` field entirely, so a `Pipeline` value is now a pure description of what to fetch with no execution state attached.

Scope notes, unchanged:

- `rawStage()`, `Expression.raw()`, `pipelineFunction()` and `PipelineFunctions.raw()` **stay** untyped — they are the documented escape hatch for preview stages.
- `search(Map<String, Object?>)` stays untyped **for now**; its own doc says the search API is still evolving, so typing it would bake in a shape likely to move. Node has no search stage at all.

---

## 7. CHANGELOG describes intra-PR history — DONE

**Type:** documentation

The `Unreleased` section's last three bullets describe changes made *within* #292 — "`Pipeline.unnest()` **now** takes…", "`replaceWith()` **now** sends…", "`minimum()`/`maximum()` are aggregate-only". Pipelines are new in this release, so there is no prior published behavior for consumers to contrast against.

**Fix:** fold them into the first bullet before `0.5.4` publishes. The new top-level exports (`greaterThan`, `greaterThanOrEqual`, `lessThan`, `lessThanOrEqual`) are also unmentioned.

---

## 9. Expression method names diverge from the Node SDK — DONE

**Type:** enhancement — public API

Found while verifying item 4. The alias removals were correct, but a full comparison against Node's live `expression.ts` (comments stripped) turned up five gratuitous name differences and four missing methods, all pre-existing on `main`.

Renamed to Node's names:

| was | now |
| --- | --- |
| `modulo` | `mod` |
| `arrayContainsElement` | `arrayContains` |
| `timestampTrunc` | `timestampTruncate` (fluent and static) |

Added fluent forms Node has and this SDK only had as statics: `stringReverse()` on `PipelineExpression`, and `not()`, `countIf()`, `conditional()` on `PipelineBooleanExpression` — matching the classes Node declares them on.

Added to the static catalog: `arrayMaximum`, `arrayMaximumN`, `arrayMinimum`, `arrayMinimumN`, `arraySum`, `countAll`. These existed as fluent methods but were missing from `PipelineFunctions`; Node exposes both forms.

**Deliberately not renamed:** fluent `toLowerCase()`/`toUpperCase()` keep Dart's `String` names rather than Node's `toLower`/`toUpper`. They are the only pair with a direct Dart-core analogue, and the wire ops are still `to_lower`/`to_upper`. This is documented on the methods so it is not "fixed" for parity later. The `PipelineFunctions` statics keep Node's names, as on `main`.

After this, the comparison is empty in both directions — every fluent method and top-level function Node exposes now exists here under the same name, except that one deliberate pair.

---

## Out of scope

`test/bundle_test.dart` (7 tests) and `test/transaction_test.dart` (1 test) fail when run locally **without the Firestore emulator**. They pass in CI, which starts the emulator. Correcting an earlier note here that called them pre-existing failures on `main` — they are not; nothing is wrong with those suites.

demolaf:
Fixes #317

Pipelines couldn't run inside a transaction at all: `execute()` read back a transaction ID from a request that never set `newTransaction`, and its `transaction:` parameter took a raw base64 string no public API could produce. Adds `Transaction.executePipeline()`, routed through the same `_withLazyStartedTransaction` path as `getQuery`, and drops the unusable parameter along with the never-populated `PipelineSnapshot.transaction`.

The rest corrects the Pipeline API against the Node Admin SDK. All of it is unpublished: `main` is version-bumped to 0.5.4 but pub.dev still serves 0.5.3, so none of this breaks anyone today.

- Renamed expression methods to Node's names: `modulo` to `mod`, `arrayContainsElement` to `arrayContains`, `timestampTrunc` to `timestampTruncate`. Added the ones Node has that were missing: fluent `stringReverse()`, `not()`, `countIf()`, `conditional()`, plus `PipelineFunctions.arrayMaximum()`, `arrayMaximumN()`, `arrayMinimum()`, `arrayMinimumN()`, `arraySum()` and `countAll()`. `toLowerCase()`/`toUpperCase()` deliberately keep Dart's `String` names.
- Aligned the result types: `PipelineResult.document` is now `ref`, matching both Node and `DocumentSnapshot.ref` elsewhere in this package, plus `PipelineResult.id`, value equality, and `PipelineSnapshot.pipeline`.
- Pipeline execution options are typed and live on `execute()` and `Transaction.executePipeline()`: `indexMode`, `explain`, and a `rawOptions` escape hatch. Requesting planner stats previously meant writing proto field names into a map, where a wrong key failed silently.
- `PipelineSnapshot.explainStats` exposed the unexported generated `firestore_v1.ExplainStats`. It is now an exported `ExplainStats` with `text` for the `TEXT`/`JSON` formats, `typeName`, and `raw` carrying the encoded payload so nothing is lost for a format we do not decode.
- `collectionReference()`, `documents()` and `union()` accepted sources from another database and silently targeted the wrong one. `_validateSameDatabase` moves off `PipelineSource` so they share it with `createFrom()`, and now compares project as well as database.
- Drops 14 methods that delegated verbatim to their base (`equalValue`, `addNumber`, and so on), plus `alias()` and `PipelineSnapshot.result`. The names come from the JS Web SDK where `eqValue` forces a constant and `eq` takes an expression; here both were identical, so `equalValue(expr)` got expression semantics. The Node Admin SDK has none of them.
- `Pipeline._execute` now retries transient connection errors like `_QueryReader` and `_AggregationReader` do, and `PipelineResult.data()` is no longer typed nullable when it never returns null.


kevmoo (commented):
> **Note:** This code review was generated and empirically red-team verified (via local unit test reproductions and upstream `googleapis/nodejs-firestore` comparisons) by an AI coding agent on behalf of @kevmoo.

### Summary
Overall this change looks very solid! All 314 unit tests and `dart analyze` checks pass locally in `packages/google_cloud_firestore`.

Also noting re: the earlier `gemini-code-assist[bot]` comment on `_validateSameDatabase` ([#3895336856](https://github.com/firebase/firebase-admin-dart/pull/318#discussion_r3895336856)): your approach in `717ed4f` using `String? get _knownProjectId` is the right call. Accessing `.projectId` synchronously before the first RPC throws a `StateError` when using async GCE metadata server / `gcloud` discovery (and upstream Node.js actually crashes with `INTERNAL ERROR: Client is not yet ready to issue requests.` in that case).

A few suggestions and minor nits below:

---

### Suggestions

1. **Wrap `Pipeline._execute` in `retryOnConnectionError` (`lib/src/pipeline.dart#L1575-L1626`)**
   Both `_QueryReader._get()` (`lib/src/query_reader.dart#L65-L128`) and `_AggregationReader._get()` (`lib/src/aggregation_reader.dart#L65-L114`) wrap their streaming RPC call in `retryOnConnectionError(..., allowRetry: transactionId == null && transactionOptions == null)` so standalone executions retry transient `http.ClientException` failures before any stream chunk arrives (while delegating retries to `Transaction._runTransaction` when inside a transaction).
   - *Implementation note*: Because a stream chunk can carry `executionTime`, `explainStats`, or `transaction` even when `chunk.results` is empty (e.g. an `analyze` explain query), track `var hasProgress = false;` (set to `true` as soon as any chunk arrives in `await for (final chunk in response)`) and reset `results.clear(); executionTime = null; newTransaction = null; explainStats = null; hasProgress = false;` at the start of the `retryOnConnectionError` closure.

2. **Self-Equality Test Tautology & `PipelineResult.operator ==` / `hashCode` (`lib/src/pipeline.dart#L1979-L1995`, `test/pipeline_test.dart#L184`)**
   - In `test/pipeline_test.dart#L184`, `expect(snapshot.results.single, equals(snapshot.results.single))` compares the exact same object instance to itself.
   - In `PipelineResult.operator ==` (`lib/src/pipeline.dart#L1979-L1985`), upstream Node.js (`dev/src/pipelines/pipelines.ts#L2066-L2072`) checks `this === other || (isOptionalEqual(this.ref, other.ref) && deepEqual(this._fieldsProto, other._fieldsProto))`. Dart's `operator ==` currently omits `identical(this, other)` and checks `other.name == name` instead of `other.ref == ref` (so two `PipelineResult` instances from two different `Firestore` instances with the same document path compare equal even though `res1.ref != res2.ref`).
   - If updated to check `other.ref == ref`, remember to also include `ref` in `PipelineResult.hashCode`.

3. **PR Description vs. `9481c7a` & Optional Execution Options on `Transaction.executePipeline` (`lib/src/transaction.dart#L164-L177`)**
   - The PR description still mentions: *"`Pipeline.withOptions()` is now typed — `withOptions(indexMode:, explain:)` instead of a map of proto field names, with `withRawOptions()` kept as the escape hatch."* (Both were removed in `9481c7a`).
   - While Node.js `Transaction.execute(pipeline: Pipeline)` takes no options, since Dart's `Pipeline._execute` already supports forwarding `options` in `StructuredPipeline`, consider adding optional `{PipelineIndexMode? indexMode, PipelineExplainOptions? explain, Map<String, Object?> rawOptions = const {}}` parameters to `Transaction.executePipeline`.

---

### Minor Nits

- **`lib/src/pipeline.dart#L132-L140`**: Docstrings on `PipelineExplainMode` and `PipelineExplainOptions.mode` still say `"plan it"` after `PipelineExplainMode.explain` was removed in `9481c7a` in favor of `execute` and `analyze`.
- **`lib/src/pipeline.dart#L2953-L2958`**: When `_validateSameDatabase` is called from `Transaction.executePipeline` (`lib/src/transaction.dart#L171`), the error message states `'... does not match the target database (...) of this Pipeline.'` instead of `'this Transaction.'`.
- **`test/transaction_test.dart#L100-L135`**: Consider adding a unit test for `tx.executePipeline(...)` to the `read-after-write guard` group alongside `get()`, `getQuery()`, and `getAll()`.

Lyokone (approved):

demolaf:
> > **Note:** This code review was generated and empirically red-team verified (via local unit test reproductions and upstream `googleapis/nodejs-firestore` comparisons) by an AI coding agent on behalf of @kevmoo.
> 
> ### Summary
> Overall this change looks very solid! All 314 unit tests and `dart analyze` checks pass locally in `packages/google_cloud_firestore`.
> 
> Also noting re: the earlier `gemini-code-assist[bot]` comment on `_validateSameDatabase` ([#3895336856](https://github.com/firebase/firebase-admin-dart/pull/318#discussion_r3895336856)): your approach in `717ed4f` using `String? get _knownProjectId` is the right call. Accessing `.projectId` synchronously before the first RPC throws a `StateError` when using async GCE metadata server / `gcloud` discovery (and upstream Node.js actually crashes with `INTERNAL ERROR: Client is not yet ready to issue requests.` in that case).
> 
> A few suggestions and minor nits below:
> 
> ### Suggestions
> 1. **Wrap `Pipeline._execute` in `retryOnConnectionError` (`lib/src/pipeline.dart#L1575-L1626`)**
>    Both `_QueryReader._get()` (`lib/src/query_reader.dart#L65-L128`) and `_AggregationReader._get()` (`lib/src/aggregation_reader.dart#L65-L114`) wrap their streaming RPC call in `retryOnConnectionError(..., allowRetry: transactionId == null && transactionOptions == null)` so standalone executions retry transient `http.ClientException` failures before any stream chunk arrives (while delegating retries to `Transaction._runTransaction` when inside a transaction).
>    
>    * _Implementation note_: Because a stream chunk can carry `executionTime`, `explainStats`, or `transaction` even when `chunk.results` is empty (e.g. an `analyze` explain query), track `var hasProgress = false;` (set to `true` as soon as any chunk arrives in `await for (final chunk in response)`) and reset `results.clear(); executionTime = null; newTransaction = null; explainStats = null; hasProgress = false;` at the start of the `retryOnConnectionError` closure.
> 2. **Self-Equality Test Tautology & `PipelineResult.operator ==` / `hashCode` (`lib/src/pipeline.dart#L1979-L1995`, `test/pipeline_test.dart#L184`)**
>    
>    * In `test/pipeline_test.dart#L184`, `expect(snapshot.results.single, equals(snapshot.results.single))` compares the exact same object instance to itself.
>    * In `PipelineResult.operator ==` (`lib/src/pipeline.dart#L1979-L1985`), upstream Node.js (`dev/src/pipelines/pipelines.ts#L2066-L2072`) checks `this === other || (isOptionalEqual(this.ref, other.ref) && deepEqual(this._fieldsProto, other._fieldsProto))`. Dart's `operator ==` currently omits `identical(this, other)` and checks `other.name == name` instead of `other.ref == ref` (so two `PipelineResult` instances from two different `Firestore` instances with the same document path compare equal even though `res1.ref != res2.ref`).
>    * If updated to check `other.ref == ref`, remember to also include `ref` in `PipelineResult.hashCode`.
> 3. **PR Description vs. `9481c7a` & Optional Execution Options on `Transaction.executePipeline` (`lib/src/transaction.dart#L164-L177`)**
>    
>    * The PR description still mentions: _"`Pipeline.withOptions()` is now typed — `withOptions(indexMode:, explain:)` instead of a map of proto field names, with `withRawOptions()` kept as the escape hatch."_ (Both were removed in `9481c7a`).
>    * While Node.js `Transaction.execute(pipeline: Pipeline)` takes no options, since Dart's `Pipeline._execute` already supports forwarding `options` in `StructuredPipeline`, consider adding optional `{PipelineIndexMode? indexMode, PipelineExplainOptions? explain, Map<String, Object?> rawOptions = const {}}` parameters to `Transaction.executePipeline`.
> 
> ### Minor Nits
> * **`lib/src/pipeline.dart#L132-L140`**: Docstrings on `PipelineExplainMode` and `PipelineExplainOptions.mode` still say `"plan it"` after `PipelineExplainMode.explain` was removed in `9481c7a` in favor of `execute` and `analyze`.
> * **`lib/src/pipeline.dart#L2953-L2958`**: When `_validateSameDatabase` is called from `Transaction.executePipeline` (`lib/src/transaction.dart#L171`), the error message states `'... does not match the target database (...) of this Pipeline.'` instead of `'this Transaction.'`.
> * **`test/transaction_test.dart#L100-L135`**: Consider adding a unit test for `tx.executePipeline(...)` to the `read-after-write guard` group alongside `get()`, `getQuery()`, and `getAll()`.

@kevmoo thanks for the feedback, I've addressed each one.

PR #322, opened 2026-09-04
fix(storage): honour AppOptions.credential instead of ADC

### LINKED ISSUE #320: `app.storage()` ignores the app `Credential` and always uses Application Default Credentials
## Summary

`Storage` builds its `google_cloud_storage.Storage` delegate without passing any HTTP client, so the GCS client falls back to Application Default Credentials (ADC) even when the `FirebaseApp` was initialized with `Credential.fromServiceAccount(...)` / `Credential.fromServiceAccountParams(...)`.

This is inconsistent with the rest of the SDK: `Firestore` (`_buildSettings`) and the app's default client (`FirebaseApp._createDefaultClient`) both honor `options.credential`.

Two consequences:

1. Storage calls are authenticated as whatever ADC resolves to (a developer's `gcloud` user, the GCE metadata server, ...) instead of the configured service account.
2. Simply calling `app.storage()` starts the ADC lookup immediately (the client `Future` is created in the `google_cloud_storage.Storage` constructor and stored in a field). When ADC is missing or expired, that `Future` fails as an **unhandled exception with no user frame in the stack trace**, even if no storage operation is ever performed.

## Environment

- `firebase_admin_sdk` 0.5.4 (also present in 0.5.1, 0.5.2, 0.5.3)
- `google_cloud_storage` 0.6.3
- `googleapis_auth` 2.3.3
- Dart SDK 3.13.2 (stable), Linux x64

## Steps to reproduce

`pubspec.yaml`:

```yaml
name: adc_repro
environment:
  sdk: ^3.9.0
dependencies:
  firebase_admin_sdk: 0.5.4
```

`bin/main.dart` (any syntactically valid service account JSON works, no network call to Firebase is made):

```dart
import 'dart:io';

import 'package:firebase_admin_sdk/firebase_admin_sdk.dart';
import 'package:firebase_admin_sdk/storage.dart';

Future<void> main() async {
  final app = FirebaseApp.initializeApp(
    options: AppOptions(
      credential: Credential.fromServiceAccount(File('service_account.json')),
      projectId: 'demo-project',
      storageBucket: 'demo-project.firebasestorage.app',
    ),
  );
  print('credential: ${app.options.credential.runtimeType}');

  // Only constructing the service, no storage call is made.
  final storage = app.storage();
  print('storage created: ${storage.runtimeType}');

  // Give the client future created in the constructor time to fail.
  await Future<void>.delayed(const Duration(seconds: 5));
  print('done');
}
```

Run it in an environment where ADC is not available (no `GOOGLE_APPLICATION_CREDENTIALS`, no `~/.config/gcloud/application_default_credentials.json`, not on GCE):

```sh
env -u GOOGLE_APPLICATION_CREDENTIALS HOME=$(mktemp -d) dart run bin/main.dart
```

## Actual result

```
credential: ServiceAccountCredential
storage created: Storage
Unhandled exception:
MetadataServerException: Could not connect to metadata.google.internal.
#0      fetchMetadataValue (package:google_cloud/src/gcp_project.dart:309:5)
<asynchronous suspension>
#1      MetadataServerAuthorizationFlow.run (package:googleapis_auth/src/oauth2_flows/metadata_server.dart:32:29)
<asynchronous suspension>
#2      clientFromFlow (package:googleapis_auth/src/oauth2_flows/base_flow.dart:31:25)
<asynchronous suspension>
#3      clientViaMetadataServer (package:googleapis_auth/src/metadata_server_client.dart:40:13)
<asynchronous suspension>
#4      clientViaApplicationDefaultCredentials (package:googleapis_auth/auth_io.dart:103:10)
<asynchronous suspension>
```

The process never prints `done`. Note that the stack trace contains no frame from `firebase_admin_sdk` or from user code, which makes this hard to diagnose.

With an **expired** gcloud ADC token (common on developer machines) the same code fails with:

```
Failed to obtain access credentials. Error: invalid_grant reauth related error (rapt_required) Status code: 400
package:googleapis_auth/src/utils.dart 123:7            ClientExtensions.requestJson
===== asynchronous gap ===========================
package:googleapis_auth/src/auth_functions.dart 175:19  refreshCredentials
===== asynchronous gap ===========================
package:googleapis_auth/src/adc_utils.dart 83:7         _clientViaApplicationCredentials
===== asynchronous gap ===========================
package:googleapis_auth/auth_io.dart 95:12              clientViaApplicationDefaultCredentials
```

In our case this surfaced as `dart test` reporting `Failed to load "test/..."` for a whole test file that only used Firestore through a service account; storage was never used, only instantiated eagerly.

## Expected result

- `app.storage()` should use the credential configured on the app, exactly like Firestore and `app.client` do. With a `ServiceAccountCredential`, no ADC lookup should happen at all.
- Constructing the service should not spawn an unawaited `Future` that can fail as an unhandled exception.

## Root cause

`lib/src/storage/storage.dart` (0.5.4):

```dart
Storage._(this.app) {
  ...
  _delegate = gcs.Storage();
}
```

`gcs.Storage()` is called without `client:` (and without `projectId:`), so `google_cloud_storage` resolves the client through its `_calculateClient(null, emulatorHost)` switch:

```dart
(null, _?) => http.Client(),        // emulator
(null, null) => defaultHttpClient(), // => clientViaApplicationDefaultCredentials(...)
```

`defaultHttpClient()` on the VM is `googleapis_auth.clientViaApplicationDefaultCredentials(scopes: [...])`, and the returned `Future` is stored in `Storage._httpClient` without being awaited or caught.

## Suggested fix

Pass the app's authenticated client (which already honors `options.credential` and is granted the `cloud-platform` scope) and the project id to the delegate. `google_cloud_storage.Storage` accepts `FutureOr<http.Client>` so no `await` is needed in the constructor:

```dart
Storage._(this.app) {
  final isEmulator = Environment.isStorageEmulatorEnabled();
  ...
  _delegate = gcs.Storage(
    projectId: app.projectId,
    // Keep the unauthenticated client in emulator mode: google_cloud_storage
    // only picks a plain http.Client when `client` is null.
    client: isEmulator ? null : app.client,
  );
}
```

This mirrors what `Firestore._buildSettings` already does for the Firestore credential and removes the implicit ADC dependency from `app.storage()`.

If a dedicated scope set is preferred over `app.client`, `app.options.credential.createClient([...devstorage scopes...])` could be used instead, falling back to ADC only when `options.credential` is an `ApplicationDefaultCredential`.

demolaf:
Closes #320

`app.storage()` built its `google_cloud_storage` delegate without a client, so Storage authenticated with Application Default Credentials even when the app was initialized with a service account, and merely calling `app.storage()` started an ADC lookup whose failure surfaced as an unhandled exception with no frame from this SDK or user code.

Storage now hands the delegate the app's authenticated client and project ID, and builds it on first use rather than in the constructor. Project ID resolution moved into a shared `FirebaseApp.resolveProjectIdSync()` so the synchronous sources — app options, environment, service account credential, credentials file — are consulted in one order; that also lets the other services resolve a configured service account's project without the gcloud CLI or metadata server hop.


kevmoo (changes requested):
> **Note:** This code review was generated and empirically red-team verified (via local unit test reproductions and upstream `firebase-admin-node` comparisons) by an AI coding agent on behalf of @kevmoo.

### Summary
Thank you for tackling #320! Lazily constructing `gcs.Storage` on first use and suppressing unhandled async errors on `_appClient()` (`client.ignore()`) is a clean way to prevent unawaited ADC network handshakes during construction.

However, empirical testing in a local worktree uncovered **2 blocking bugs** that break custom `AppOptions.httpClient` lifecycles and cross-project service account credentials in production:

---

### 🚨 Blocking Issues

1. **`Storage.delete()` unconditionally closes `app.client` (`packages/firebase_admin_sdk/lib/src/storage/storage.dart#L149-L153`)**
   - In `package:google_cloud_storage` (`src/client.dart#L181-L195`), `Storage.close()` unconditionally closes its underlying `_httpClient` (both synchronously via `_cachedServiceClient.close()` -> `client.close()` if used, and asynchronously via `future.then((client) => client.close())` if unused).
   - Because PR #322 now passes `app.client` to `gcs.Storage` in production mode (`client: _isEmulator ? null : _appClient()`), calling `Storage.delete()` (including during `FirebaseApp.close()`) closes `app.client`:
     - **Breaks User-Provided `AppOptions.httpClient`**: `FirebaseApp.client` (`firebase_app.dart#L132-L135`) and `FirebaseApp.close()` (`firebase_app.dart#L330-L332`) explicitly contract that user-provided `options.httpClient` instances are never closed by `FirebaseApp`. Adding a test in `storage_test.dart` with `AppOptions(projectId: projectId, httpClient: mockClient)`, calling `storage.bucket('some-bucket')`, and calling `await app.close()` fails immediately with `Unexpected calls: MockAuthClient.close()`. (Existing tests missed this because `mocktail` returns `null` on unstubbed `void` methods unless `verifyNever(() => mockClient.close())` is asserted.)
     - **Inconsistent with `Auth.delete()` and `Functions.delete()`**: Both `Auth.delete()` (`auth.dart#L53`) and `Functions.delete()` (`functions.dart#L99`) only close the HTTP client when running in emulator mode (`if (Environment.isAuthEmulatorEnabled())`), leaving `app.client` lifecycle management to `FirebaseApp.close()`.
   - **Suggested Fix**:
     ```dart
     @override
     Future<void> delete() async {
       if (_isDeleted) return;
       _isDeleted = true;
       // In emulator mode, gcs.Storage creates its own unauthenticated http.Client
       // which must be closed here. In production mode, it uses app.client which is
       // managed and closed by FirebaseApp.close().
       if (_isEmulator) {
         _delegateOrNull?.close();
       }
     }
     ```

2. **`resolveProjectIdSync()` checks ambient `Platform.environment` before `AppOptions.credential` in production (`packages/firebase_admin_sdk/lib/src/app/firebase_app.dart#L198-L206`)**
   - In commit `dec0d60` (`fix(storage): make project ID tests independent of the ambient environment`), `resolveProjectIdSync()` was written as:
     ```dart
     final credentialProjectId =
         options.credential?.serviceAccountCredentials?.projectId;
     if (injectedEnv != null) return credentialProjectId;

     final processProjectId = _projectIdFromEnvironment(Platform.environment);
     if (processProjectId != null) return processProjectId;

     return credentialProjectId ?? google_cloud.projectIdFromCredentialsFile();
     ```
   - Because `if (injectedEnv != null) return credentialProjectId;` only triggers when `injectedEnv != null` (i.e. inside unit tests!), **in real production execution (`injectedEnv == null`), `_projectIdFromEnvironment(Platform.environment)` is still evaluated before `credentialProjectId`!**
   - **Violates Upstream Node.js Admin SDK & PR CHANGELOG**:
     - In `firebase/firebase-admin-node` (`getExplicitProjectId` in `src/utils/index.ts`), `options.credential.projectId` (`ServiceAccountCredential`) is checked immediately after `options.projectId` and **before** `process.env.GOOGLE_CLOUD_PROJECT || process.env.GCLOUD_PROJECT`.
     - On Cloud Run, Cloud Functions, and App Engine, `GOOGLE_CLOUD_PROJECT` is set in `Platform.environment` to the **host project**. Thus in production (`injectedEnv == null`), `resolveProjectIdSync()` returns the host project instead of the service account's `project_id`, contradicting the CHANGELOG entry (*"Apps configured with a service account for one project while running on Google Cloud infrastructure in another now resolve the service account's project instead of the host project."*).
   - **Suggested Fix** (also swapping items 4 and 5 in the dartdoc list at lines 168-174):
     ```dart
     @internal
     String? resolveProjectIdSync({
       String? projectIdOverride,
       Map<String, String>? environment,
     }) {
       final injectedEnv =
           environment ?? Zone.current[envSymbol] as Map<String, String>?;
       if (injectedEnv != null) {
         final injected = _projectIdFromEnvironment(injectedEnv);
         if (injected != null) return injected;
       }

       final explicitProjectId = projectIdOverride ?? options.projectId;
       if (explicitProjectId != null) return explicitProjectId;

       final credentialProjectId =
           options.credential?.serviceAccountCredentials?.projectId;
       if (credentialProjectId != null) return credentialProjectId;

       if (injectedEnv != null) return null;

       final processProjectId = _projectIdFromEnvironment(Platform.environment);
       if (processProjectId != null) return processProjectId;

       return google_cloud.projectIdFromCredentialsFile();
     }
     ```

---

### 💡 Suggestions

1. **Update `Firestore._buildSettings` to use `app.resolveProjectIdSync()` (`packages/firebase_admin_sdk/lib/src/firestore/firestore.dart#L121`)**
   - `Firestore._buildSettings` still assigns `final projectId = app.projectId;` (`options.projectId` only). When an app is initialized with `AppOptions(credential: Credential.fromServiceAccount(saFile))` without explicit `options.projectId` or env vars, calling `app.firestore().collection('users')` throws `StateError: Project ID has not been discovered yet...` because `_settings.projectId` is `null` and `FirestoreHttpClient.getProjectId()` never checks `credential.serviceAccountCredentials?.projectId`.
   - Changing line 121 of `firestore.dart` to `final projectId = app.resolveProjectIdSync();` fixes `app.firestore().collection('users')`.
   - *(Note: For full cross-project Firestore parity when `GOOGLE_CLOUD_PROJECT` is set in `Platform.environment`, `FirestoreHttpClient._run()` in `package:google_cloud_firestore` should also check `_settings.projectId` before `getProjectId()`).*

2. **Add Regression Unit Tests (`firebase_app_test.dart`, `storage_test.dart`)**
   - Add a unit test verifying that `options.credential` (`ServiceAccountCredential`) takes precedence over ambient `GOOGLE_CLOUD_PROJECT` / `GCLOUD_PROJECT` environment variables when `options.projectId` is omitted.
   - Add a unit test in `storage_test.dart` verifying `verifyNever(() => mockClient.close())` when `AppOptions(httpClient: mockClient)` is passed and `app.close()` is called after `storage.bucket()`.

demolaf:
> > **Note:** This code review was generated and empirically red-team verified (via local unit test reproductions and upstream `firebase-admin-node` comparisons) by an AI coding agent on behalf of @kevmoo.
> 
> ### Summary
> Thank you for tackling #320! Lazily constructing `gcs.Storage` on first use and suppressing unhandled async errors on `_appClient()` (`client.ignore()`) is a clean way to prevent unawaited ADC network handshakes during construction.
> 
> However, empirical testing in a local worktree uncovered **2 blocking bugs** that break custom `AppOptions.httpClient` lifecycles and cross-project service account credentials in production:
> 
> ### 🚨 Blocking Issues
> 1. **`Storage.delete()` unconditionally closes `app.client` (`packages/firebase_admin_sdk/lib/src/storage/storage.dart#L149-L153`)**
>    
>    * In `package:google_cloud_storage` (`src/client.dart#L181-L195`), `Storage.close()` unconditionally closes its underlying `_httpClient` (both synchronously via `_cachedServiceClient.close()` -> `client.close()` if used, and asynchronously via `future.then((client) => client.close())` if unused).
>    * Because PR #322 now passes `app.client` to `gcs.Storage` in production mode (`client: _isEmulator ? null : _appClient()`), calling `Storage.delete()` (including during `FirebaseApp.close()`) closes `app.client`:
>      
>      * **Breaks User-Provided `AppOptions.httpClient`**: `FirebaseApp.client` (`firebase_app.dart#L132-L135`) and `FirebaseApp.close()` (`firebase_app.dart#L330-L332`) explicitly contract that user-provided `options.httpClient` instances are never closed by `FirebaseApp`. Adding a test in `storage_test.dart` with `AppOptions(projectId: projectId, httpClient: mockClient)`, calling `storage.bucket('some-bucket')`, and calling `await app.close()` fails immediately with `Unexpected calls: MockAuthClient.close()`. (Existing tests missed this because `mocktail` returns `null` on unstubbed `void` methods unless `verifyNever(() => mockClient.close())` is asserted.)
>      * **Inconsistent with `Auth.delete()` and `Functions.delete()`**: Both `Auth.delete()` (`auth.dart#L53`) and `Functions.delete()` (`functions.dart#L99`) only close the HTTP client when running in emulator mode (`if (Environment.isAuthEmulatorEnabled())`), leaving `app.client` lifecycle management to `FirebaseApp.close()`.
>    * **Suggested Fix**:
>      ```dart
>      @override
>      Future<void> delete() async {
>        if (_isDeleted) return;
>        _isDeleted = true;
>        // In emulator mode, gcs.Storage creates its own unauthenticated http.Client
>        // which must be closed here. In production mode, it uses app.client which is
>        // managed and closed by FirebaseApp.close().
>        if (_isEmulator) {
>          _delegateOrNull?.close();
>        }
>      }
>      ```
> 2. **`resolveProjectIdSync()` checks ambient `Platform.environment` before `AppOptions.credential` in production (`packages/firebase_admin_sdk/lib/src/app/firebase_app.dart#L198-L206`)**
>    
>    * In commit `dec0d60` (`fix(storage): make project ID tests independent of the ambient environment`), `resolveProjectIdSync()` was written as:
>      ```dart
>      final credentialProjectId =
>          options.credential?.serviceAccountCredentials?.projectId;
>      if (injectedEnv != null) return credentialProjectId;
>      
>      final processProjectId = _projectIdFromEnvironment(Platform.environment);
>      if (processProjectId != null) return processProjectId;
>      
>      return credentialProjectId ?? google_cloud.projectIdFromCredentialsFile();
>      ```
>    * Because `if (injectedEnv != null) return credentialProjectId;` only triggers when `injectedEnv != null` (i.e. inside unit tests!), **in real production execution (`injectedEnv == null`), `_projectIdFromEnvironment(Platform.environment)` is still evaluated before `credentialProjectId`!**
>    * **Violates Upstream Node.js Admin SDK & PR CHANGELOG**:
>      
>      * In `firebase/firebase-admin-node` (`getExplicitProjectId` in `src/utils/index.ts`), `options.credential.projectId` (`ServiceAccountCredential`) is checked immediately after `options.projectId` and **before** `process.env.GOOGLE_CLOUD_PROJECT || process.env.GCLOUD_PROJECT`.
>      * On Cloud Run, Cloud Functions, and App Engine, `GOOGLE_CLOUD_PROJECT` is set in `Platform.environment` to the **host project**. Thus in production (`injectedEnv == null`), `resolveProjectIdSync()` returns the host project instead of the service account's `project_id`, contradicting the CHANGELOG entry (_"Apps configured with a service account for one project while running on Google Cloud infrastructure in another now resolve the service account's project instead of the host project."_).
>    * **Suggested Fix** (also swapping items 4 and 5 in the dartdoc list at lines 168-174):
>      ```dart
>      @internal
>      String? resolveProjectIdSync({
>        String? projectIdOverride,
>        Map<String, String>? environment,
>      }) {
>        final injectedEnv =
>            environment ?? Zone.current[envSymbol] as Map<String, String>?;
>        if (injectedEnv != null) {
>          final injected = _projectIdFromEnvironment(injectedEnv);
>          if (injected != null) return injected;
>        }
>      
>        final explicitProjectId = projectIdOverride ?? options.projectId;
>        if (explicitProjectId != null) return explicitProjectId;
>      
>        final credentialProjectId =
>            options.credential?.serviceAccountCredentials?.projectId;
>        if (credentialProjectId != null) return credentialProjectId;
>      
>        if (injectedEnv != null) return null;
>      
>        final processProjectId = _projectIdFromEnvironment(Platform.environment);
>        if (processProjectId != null) return processProjectId;
>      
>        return google_cloud.projectIdFromCredentialsFile();
>      }
>      ```
> 
> ### 💡 Suggestions
> 1. **Update `Firestore._buildSettings` to use `app.resolveProjectIdSync()` (`packages/firebase_admin_sdk/lib/src/firestore/firestore.dart#L121`)**
>    
>    * `Firestore._buildSettings` still assigns `final projectId = app.projectId;` (`options.projectId` only). When an app is initialized with `AppOptions(credential: Credential.fromServiceAccount(saFile))` without explicit `options.projectId` or env vars, calling `app.firestore().collection('users')` throws `StateError: Project ID has not been discovered yet...` because `_settings.projectId` is `null` and `FirestoreHttpClient.getProjectId()` never checks `credential.serviceAccountCredentials?.projectId`.
>    * Changing line 121 of `firestore.dart` to `final projectId = app.resolveProjectIdSync();` fixes `app.firestore().collection('users')`.
>    * _(Note: For full cross-project Firestore parity when `GOOGLE_CLOUD_PROJECT` is set in `Platform.environment`, `FirestoreHttpClient._run()` in `package:google_cloud_firestore` should also check `_settings.projectId` before `getProjectId()`)._
> 2. **Add Regression Unit Tests (`firebase_app_test.dart`, `storage_test.dart`)**
>    
>    * Add a unit test verifying that `options.credential` (`ServiceAccountCredential`) takes precedence over ambient `GOOGLE_CLOUD_PROJECT` / `GCLOUD_PROJECT` environment variables when `options.projectId` is omitted.
>    * Add a unit test in `storage_test.dart` verifying `verifyNever(() => mockClient.close())` when `AppOptions(httpClient: mockClient)` is passed and `app.close()` is called after `storage.bucket()`.

@kevmoo Both blockers are fixed: `Storage.delete()` now only closes the delegate in emulator mode, and `resolveProjectIdSync()` checks the credential before the ambient environment, matching `getExplicitProjectId` upstream. Both suggestions are in as well, including `Firestore._buildSettings`, each with a regression test that fails on the old code, and the branch is rebased onto main.

kevmoo (approved):

